PR #1, opened 2024-03-09
Refactor: Buildscripts

emyfops:
This pull request changes few things in the buildscripts

- `project.properties` were changed to `property(...)`
- You can now include the dependencies jar files directly into the jar file using `includeMod` and `includeLib`


build.gradle.kts
```diff
@@ -64,21 +65,21 @@ allprojects {
...
 
-    tasks {
-        withType<JavaCompile> {
-            options.encoding = "UTF-8"
-            options.release = 17
-        }
-        compileKotlin {
-            kotlinOptions.jvmTarget = "17"
+        flatDir {
+            dirs("libs") // TODO: Absolute path
```
emyfops: Is it possible to have a absolute flat dir path ?

common/build.gradle.kts
```diff
@@ -21,11 +21,15 @@ dependencies {
     // We depend on fabric loader here to use the fabric @Environment annotations and get the mixin dependencies
     // Do NOT use other classes from fabric loader
     modImplementation("net.fabricmc:fabric-loader:$fabricLoaderVersion")
+
     // Add dependencies on the required Kotlin modules.
     modImplementation("net.fabricmc:fabric-language-kotlin:$fabricKotlinVersion")
     implementation(annotationProcessor("io.github.llamalad7:mixinextras-common:$mixinExtrasVersion")!!)
+
+    // Baritone
```
emyfops: Do we add baritone right now or we wait later

fabric/build.gradle.kts
```diff
@@ -20,16 +20,39 @@ val common: Configuration by configurations.creating {
...
+        include(it)
     }
-    modImplementation("net.fabricmc:fabric-loader:$fabricLoaderVersion")
-    modImplementation("net.fabricmc.fabric-api:fabric-api:$fabricApiVersion")
-    modImplementation("net.fabricmc:fabric-language-kotlin:$fabricKotlinVersion")
+}
+
+dependencies {
+    // Fabric API
+    includeMod("net.fabricmc:fabric-loader:$fabricLoaderVersion")
```
emyfops: If you don't want to include the mod, for example to avoid mod conflict, then only use modImplementation

build.gradle.kts
emyfops: Fix dbaadbcfbd2e21b5916ed619143b9b41cb3fbcce

fabric/build.gradle.kts
emyfops: Decided that it would be better to not include to because we want to allow multi versions
https://github.com/Avanatiker/NeoLambda/commit/737e2918ab64df9467c3380fb2185d1fbb22d8b0

PR #9, opened 2024-04-19
Feature: Fast break

emyfops:
Self explanatory

We could remove the ghost block client side


common/src/main/kotlin/com/lambda/module/modules/player/FastPlace.kt
```diff
@@ -0,0 +1,20 @@
...
+object FastPlace : Module(
+    name = "FastPlace",
+    description = "Place blocks faster.",
+    defaultTags = setOf(
+        ModuleTag.PLAYER, ModuleTag.WORLD
+    )
+) {
+    init {
+        listener<TickEvent.Pre> {
+            mc.itemUseCooldown = 0
```
emyfops: We can probably use this for other cooldowns

PR #33, opened 2024-07-08
[1.20.x] [All] Support major versions

emyfops:
This pull request modifies the mod's file dependency to be in range of the major version of minecraft since modloaders assure compatibility for all major versions

However, thedarkcolour/KotlinForForge could cause issues due to inconsistent version releasing and compatibility

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -28,7 +28,7 @@
   "depends": {
     "fabricloader": ">=${fabricLoaderVersion}",
     "fabric-api": ">=${fabricApiVersion}+${minecraftVersion}",
-    "minecraft": ">=${minecraftVersion}",
+    "minecraft": "~${minecraftVersion}",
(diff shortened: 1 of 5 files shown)


emyfops:
We could drop the use of the library for forge-like and simply transform the entrypoint objects into classes so the jvm doesn't panic when it sees a private constructor

emyfops:
After some tests and investigations kff will not cause us any issues unlike Baritone which has not only a broken maven publish but also only one release per major mc version

emyfops:
## Tests

- Forge 1.20.3/1.20.4 49.0.1-49.0.3 ✅  but _PrismLauncher/PrismLauncher/issues/1934_
- Forge 1.20.4-49.0.1 ✅ 
- Forge 1.20.4-49.0.21 ✅ 
- Forge 1.20.4-49.1.0 ✅ 
- Forge 1.20.4-49.1.4 ✅ 


For some reasons Fabric simply don't boot up, nothing related to the mod but I know it works just fine
I was able to launch the mod using NeoForge on any version but #34 prevented me to actually play

PR #60, opened 2024-08-20
[1.20.4] Feat: Skybox and fog colors

Camii2407:


common/src/main/java/com/lambda/mixin/render/RenderSystemMixin.java
```diff
@@ -0,0 +1,26 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(RenderSystem.class)
+public class RenderSystemMixin {
+    @Shadow @Final private static float[] shaderFogColor;
+
+    @Inject(method = "_setShaderFogColor", at = @At(value ="HEAD"), cancellable = true)
```
emyfops, line 16: Is it possible to chain multiple `@ModifyArg` annotations to achieve the same result ?
Don't use the `@ModifyArgs` one, forge doesn't like it

emyfops (approved):

common/src/main/java/com/lambda/mixin/render/RenderSystemMixin.java
Avanatiker: 1 mixin better than 3?

emyfops: Yes

Avanatiker: Also inject is not invasive

common/src/main/kotlin/com/lambda/util/math/ColorUtils.kt
```diff
@@ -13,4 +14,7 @@ object ColorUtils {
     val Color.g get() = green.toDouble() / 255.0
     val Color.b get() = blue.toDouble() / 255.0
     val Color.a get() = alpha.toDouble() / 255.0
+
+    @JvmStatic
+    val Color.vec3d get() = Vec3d(r, g, b)
```
emyfops, line 19: The extension name seems counter-intuitive
What's the purpose of having a vector of color ?

Avanatiker: It makes no sense but MC uses Vec3d instead of Color xD

ToxicAven: The main thing is that in ColorUtils it is already being cast to double, might as well take advantage and not repeat the division in the mixin :P

Avanatiker: its so funny that we made the same utility function except that ours was a field extension and yours was a function :D

ToxicAven: What? That was mine

Avanatiker: nope yours was `fun Color.toVec3d() = Vec3d(r, g, b)` and we did
```
@JvmStatic
val Color.vec3d get() = Vec3d(r, g, b)
```
a few days ago

PR #63, opened 2024-08-24
[1.20.4] Ref: Various GL optimization

emyfops:
# Description

This pull request utilize few techniques to speed up the initial loading process by:
- Adding a cache or compute system that can be used anywhere for I/O
- Includes a much better PNG decoder for writing data
The game now launches 2x to 3x faster

Additionally, with the new reflections utilities, all reflections logging can be displayed at the top of the console, preventing it from flooding our initial logging.
However, this comes with the trade-off of excluding measureTimeMillisecond in each phase, except the final one
(Personally, I prefer it this way :trollface:)

This pull request also add Blade's game loop event

Please test this pr before merging


common/src/main/kotlin/com/lambda/module/modules/client/RenderSettings.kt
```diff
@@ -18,6 +18,10 @@ object RenderSettings : Module(
     val baselineOffset by setting("Vertical Offset", 0.0, -10.0..10.0, 0.5) { page == Page.Font }
     private val lodBiasSetting by setting("Smoothing", 0.0, -10.0..10.0, 0.5) { page == Page.Font }
 
+    // Texture
```
Avanatiker: we really have to move many of those options to a global settings view

Avanatiker (approved):

common/src/main/kotlin/com/lambda/module/modules/client/RenderSettings.kt
emyfops: These are actually settings I forgot to remove, can you do it ?

Avanatiker: ok bro

Avanatiker:
Very cool

PR #104, opened 2025-02-14
[1.20.4] Ref: Inline setting event functions

emyfops:
This pull request changes the `AbstractSetting` event subscribers such as `onValueChanged`, `onValueSet` and others to facilitate subscription without having to wrap it in an `apply block`

Before:
```kt
private val mesh by setting("Mesh", true, "Connect similar adjacent blocks").apply { onValueSet(::rebuildMesh) }
```

After
```kt
private val mesh by setting("Mesh", true, "Connect similar adjacent blocks").onValueSet(::rebuildMesh)
```

Instead, we call `apply` in the event subscription event
```kt
fun onValueSet(block: (from: T, to: T) -> Unit) = apply {
    listeners.add(ValueListener(false, block))
}
```


emyfops:
Tested, working

PR #122, opened 2025-05-21
Feat: Add click packet limit to PacketLimiter

IceTank:
### Description
This feature adds a packet limit tracker for click packets using a sliding window rate limit model. When configured, inventory click actions exceeding the limit are canceled. The feature has options in the module to enable, disable and change feature parameters.

### Why this feature matters
This module helps you prevent getting kicked on 2b by canceling inventory click actions that would get you kicked for click packet spam.

### Checklist
- All tests pass
- This feature is very good
- Code quality is probably ok idk how kotlin works
- pls merge


common/src/main/kotlin/com/lambda/event/events/RenderEvent.kt
```diff
@@ -40,6 +42,9 @@ sealed class RenderEvent {
         class Scaled(scaleFactor: Double) : GUI(scaleFactor)
         class HUD(scaleFactor: Double) : GUI(scaleFactor)
         class Fixed : GUI(1.0)
+        class Container(val genericContainerScreen: GenericContainerScreen, val drawContext: DrawContext, val mouseX: Int, val mouseY: Int, val delta: Float) : GUI(1.0) {
+            val mouse = Vec2d(mouseX, mouseY)
+        }
```
emyfops: We already have container events and mouse events, this is not needed

common/src/main/kotlin/com/lambda/module/modules/network/PacketLimiter.kt
```diff
@@ -58,11 +77,46 @@ object PacketLimiter : Module(
...
+            val renderScreen: HandledScreen<*> = it.genericContainerScreen
+            val context: DrawContext = it.drawContext
+            val x = renderScreen.x
+            val y = renderScreen.y
+
+            RenderSystem.disableDepthTest()
+            val remainingText = "Clicks Remaining: " + clickPacketsRemaining().toInt().toString()
+            context.drawText(renderScreen.textRenderer, Text.literal(remainingText), x + renderScreen.backgroundWidth, y, 4210752, false)
+            RenderSystem.enableDepthTest()
+        }
```
emyfops: This should be replaced by RenderEvent.GUI.Scaled and then use `FontRenderer` to build the string information that will then be dispatched to opengl

You should also get the current screen with the global context `mc` property of the SafeContext class, every extensions of that class has access to its properties and other extensions
This allows for a safe access of the game's content with the expectation of these properties to not be null

You can read more about that at https://kotlinlang.org/docs/extensions.html#

common/src/main/kotlin/com/lambda/module/modules/network/PacketLimiter.kt
```diff
@@ -58,11 +77,46 @@ object PacketLimiter : Module(
...
     }
+
+    fun canSendClickPackets(packets: Int) = clickPacketQueue.size + packets < clickPacketsWindowAmount()
+    fun clickPacketsRemaining() = clickPacketsWindowAmount() - clickPacketQueue.size
+
+    private fun clickPacketsWindowAmount() = limitClickWindowSize * limitClickRate
```
emyfops: Simple methods returning values should be replaced by properties with getters

https://kotlinlang.org/docs/coding-conventions.html#properties

common/src/main/java/com/lambda/mixin/render/GenericContainerScreenMixin.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.client.gui.screen.ingame.GenericContainerScreen;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+/*
+ * @author IceTank
+ * @since 21.05.2025
+ */
```
emyfops: We have `git blame` in intellij 😄

emyfops (changes requested):

common/src/main/kotlin/com/lambda/event/events/RenderEvent.kt
IceTank: I need access to HandledScreen render methods which I did not find any existing mixins for. x y and background width don't exist on Screen.

common/src/main/kotlin/com/lambda/module/modules/network/PacketLimiter.kt
IceTank: I can look into that

common/src/main/kotlin/com/lambda/module/modules/network/PacketLimiter.kt
IceTank: Ok

common/src/main/java/com/lambda/mixin/render/GenericContainerScreenMixin.java
IceTank: I can remove the author templates. I took this idea from the baritone project.

common/src/main/kotlin/com/lambda/event/events/RenderEvent.kt
emyfops: This comment goes alongside the module's event review

common/src/main/java/com/lambda/mixin/render/GenericContainerScreenMixin.java
IceTank: File was removed

IceTank:
@Edouard127 pls review

emyfops (approved):

Avanatiker (changes requested):
Thanks for your first contribution! :) for me the click counter does not show ingame. so we need to rework this. does it work for you guys?

IceTank:
> Thanks for your first contribution! :) for me the click counter does not show ingame. so we need to rework this. does it work for you guys?

I just tested it, and it shows up for me. There is a setting in the module to disable it showing in guis. It should be at the top of chests.

PR #124, opened 2025-05-23
feat: flight module

mmvanheusden:
### Description
Add various flight modules.  
My vision is to add various (but not complex) flight modes, so everyone can choose the mode they prefer.

### DIFF
--- a/common/src/main/kotlin/com/lambda/module/modules/movement/Flight.kt
+++ b/common/src/main/kotlin/com/lambda/module/modules/movement/Flight.kt
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2025 Lambda
+ *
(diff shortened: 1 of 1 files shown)


emyfops:
Does with work with grim 2 & 3 ?

mmvanheusden:
> Does with work with grim 2 & 3 ?

No :crying_cat_face: just creativefly for now

emyfops:
Make this pr a draft then, we had a flight module before and removed it because it didn't work on grim anymore

mmvanheusden:
> Make this pr a draft then

Blame IDEA

> removed it because it didn't work on grim anymore

Does the client exclusively have to be Grim bypasses?

beanbag44:
just tested it, there are still some debug statements left in the module btw

PR #125, opened 2025-05-27
Feature/antilevitation

mmvanheusden:
### Description
Adds LevitationTweaks, with a `REMOVE` and `UNCP` mode.
`UNCP` mode is designed to by default fully abuse poor checks in Updated-NoCheatPlus and might also work in legacy NCP (not tested).
The `REMOVE` mode has a unique feature that reapplies the effect when the module is disabled before the (server-sided) levitation effect runs out, as most clients just remove the effect which causes desynchronization. It is currently implemented using a time comparison, but on servers with poor TPS this may cause issues so this needs to be looked at

Code is functional but not yet refactored

Showcase:
https://youtu.be/Vi9CGH3xkBU

### DIFF
--- a/common/src/main/kotlin/com/lambda/module/modules/movement/LevitationTweaks.kt
+++ b/common/src/main/kotlin/com/lambda/module/modules/movement/LevitationTweaks.kt
@@ -48,26 +48,17 @@ object LevitationTweaks : Module(
     private val restore by setting("Restore", true, description = "Restore levitation on module disable") { mode == Mode.REMOVE }
 
     // UNCP
-    private val constantUpFactor by setting(
-        "Idle Up Speed", unit = "%", description = "Permanent upwards motion", defaultValue = 2.0, range = 0.0..100.0
-    ) { mode == Mode.UNCP}
-    private val control by setting("Control", true, ) { mode == Mode.UNCP}
-    private val strafeBoost by setting("Strafe Boost", true, ) { mode == Mode.UNCP && control }
-    private val strafeBoostSpeed by setting(
-        "Boost Speed", unit = "%", defaultValue = 100, range = 0..120
-    ) { mode == Mode.UNCP && control && strafeBoost}
-    private val timer by setting(
-        "Timer", true
-    ) { mode == Mode.UNCP && control && strafeBoost}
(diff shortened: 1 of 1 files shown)


common/src/main/kotlin/com/lambda/module/modules/movement/LevitationTweaks.kt
```diff
@@ -0,0 +1,158 @@
...
+        "Timer", true
+    ) { mode == Mode.UNCP && control && strafeBoost}
+    private val timerBoost by setting("Timer Boost", 1.08, 1.0..1.2, 0.01) { mode == Mode.UNCP && control  && timer}
+
+    private val controlDownSpeed by setting(
+        "Control Down Speed", unit = "%", defaultValue = 100, range = 0..300
+    ) { mode == Mode.UNCP && control}
+    private val controlUpSpeed by setting(
+        "Control Up Speed", unit = "%", defaultValue = 100, range = 0..140
+    ) { mode == Mode.UNCP && control}
```
emyfops: Settings arguments should be of the following order:
- Name
- Default value
- Range (if applicable)
- Step (if applicable)
- Description
- Unit
- Filter

And should always be inlined, only setting events can be placed underneath the value definition 
This is to have a concise setting format throughout the whole codebase

common/src/main/kotlin/com/lambda/module/modules/movement/LevitationTweaks.kt
```diff
@@ -0,0 +1,158 @@
...
+
+
+
+    private enum class Mode(override val displayName: String) : NamedEnum {
+        REMOVE("Remove"), UNCP("NCP New")
+    }
+
+    private var capturedEffect: StatusEffectInstance? = null
```
emyfops, line 68: This should be a lazy updatable that updates the effect on demand to avoid mutating states, see the utils

common/src/main/kotlin/com/lambda/module/modules/movement/LevitationTweaks.kt
```diff
@@ -0,0 +1,158 @@
...
+    private var wearOffTime: LocalDateTime? = null
+    private var canMove = false
```
emyfops, line 70: Maybe this could be a getter

PR #126, opened 2025-05-30
Added kotlin `Duration` settings

emyfops:
### Description
- Adds new duration type setting that takes in a base duration value, a duration range and a step value.
- Adds a duration highest unit inferrer
- Changes the `LimitedDecayQueue` to use monotonic time marks
- Changes the `NumericSettings` class to allow any comparable types

These changes are recommended because comparing longs and ints for duration checks are prone to errors and using durations and monotonic time marks makes it easier to compare durations


Avanatiker (changes requested):
Good start but delay setting is not yet shown in the ClickGUI and also we need some kind of option that allows to render the duration as time or tick based depending on the usecase ticks are more relevant. maybe a wrapper function that also automatically has 50ms as step idk

emyfops:
This pull request is outdated and needs to be updated for ImGui, do not delete this branch

PR #127, opened 2025-06-25
Feature: FreeLook Module

IceTank:
FreeLook Module

Has some extra settings for more advanced use


common/src/main/kotlin/com/lambda/module/modules/render/FreeLook.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    @JvmStatic
+    fun updateCam() {
+        Lambda.mc.gameRenderer.apply {
+            camera.setRotation(FreeLook.camera.yawF, FreeLook.camera.pitchF)
+        }
+    }
+
+    init {
+        previousPerspective = MinecraftClient.getInstance().options.perspective
```
emyfops: You might as well use the `mc` property in `Lambda`

common/src/main/kotlin/com/lambda/module/modules/render/FreeLook.kt
```diff
@@ -0,0 +1,82 @@
...
+import com.lambda.interaction.request.rotation.Rotation
+import com.lambda.module.Module
+import com.lambda.util.extension.rotation
+import net.minecraft.client.MinecraftClient
+import net.minecraft.client.option.Perspective
+
+object FreeLook : Module(
+    name = "FreeLook",
+    description = "Allows you to look around freely while moving",
+    defaultTags = setOf(com.lambda.module.tag.ModuleTag.RENDER, com.lambda.module.tag.ModuleTag.MOVEMENT)
```
emyfops: You can import the enum instead of using the full canonical path

emyfops (approved):

common/src/main/java/com/lambda/mixin/render/CameraMixin.java
```diff
@@ -93,8 +98,18 @@ private void onClipToSpace(float desiredCameraDistance, CallbackInfoReturnable<F
...
         return desiredCameraDistance;
     }
+
+    @ModifyArgs(method = "update", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/Camera;setRotation(FF)V"))
+    private void onUpdateSetRotationArgs(Args args) {
+        if (FreeLook.INSTANCE.isEnabled()) {
+            args.set(0, FreeLook.INSTANCE.getCamera().getYawF());
+            args.set(1, FreeLook.INSTANCE.getCamera().getPitchF());
+        }
+    }
```
emyfops: Could you add a snippet of the modified method ? Check other mixins for an example 
This is crucial for mixin updates on new versions

emyfops (changes requested):

common/src/main/java/com/lambda/mixin/render/CameraMixin.java
IceTank: Do you mean the original source?

emyfops: Yes, look at other mixins for an example

IceTank: Will do

emyfops (approved):
lgtm @Avanatiker what do you think

IceTank:
Btw this merges into 1.21.5. But I think 1.21.4 is main atm? Can probably back port easily.

emyfops:
> Btw this merges into 1.21.5. But I think 1.21.4 is main atm? Can probably back port easily.

1.21.5 is main

PR #128, opened 2025-07-17
Feature: ViewModel module

beanbag44:
This module adds a way to modify the vanilla rendering of held items.

You can do / adjust:
- scale
- position 
- rotation
- field of view
- constant swing progress
- disable / enable hand rendering adjustments
- split changes between both hands
- modify which hand is swinging
- use old minecraft swing animations
- remove the bobbing effect when switching items
- remove the delay between swinging animations

There is one todo left that i want to achieve at some point but doesn't seem worth it at the moment. Its possible to remove the shadows from the rendered items however i haven't figured out how to implement that yet.


emyfops (approved):
looks good, just some nits i want to do

PR #131, opened 2025-07-29
Improvement: use minecraft's keybind structure for inventory move

beanbag44:
alters the inventory move module to use the players keybinds over default, and lets minecraft handle the movement logic


emyfops:
This is a regression as the user is no longer able to rotate using the arrow keys

beanbag44:
oh sorry i didn't mean to remove that feature. I'll fix it. I don't think any of the changes i made should affect it

Avanatiker:
Thanks but i had to redo it from scratch due to massively outdated target branch :/

PR #133, opened 2025-08-04
Change default font size to 18 pixels

beanbag44:
This allows the user to scale down the size if wanted, rather than scaling up and losing quality


Avanatiker:
I will make a font atlas rebuilder with variable text size selector. also iam working on a scale option

PR #135, opened 2025-08-14
Fix: Fixes the main compatibility issue in the break manager after the 1.21.5 update

beanbag44:
Properly propagates and tracks values to make sure blocks are broken within the correct time frame, accounting for the server players attribute update delays


Avanatiker:
Sorry i did a refactor on whole master so some conflicts arose ;-;

Avanatiker (approved):

PR #147, opened 2025-09-20
Feature: AutoEat and Building Life Maintainance

Avanatiker:
Adds a EatTask with smort behaviour and tons of settings used by both BuildTask and AutoEat.


emyfops (approved):
Needs an option to bulk but looks good to me

Avanatiker:
What do you mean with bulk?

emyfops:
> What do you mean with bulk?
> 

Muscles building by bulking :3

PR #150, opened 2025-09-24
Feature: Mouse buttons in keybinds

beanbag44:
Implements logic for handling mouse button inputs for keybinds


emyfops (approved):
We could also combine keys and mouse for bind combination. I.e ctrl + left click + a, shift + c, etc.

beanbag44:
I wondered about doing that too, but figured i'd implement that after these changes are reviewed and verified

PR #154, opened 2025-09-28
Feature (Module): Add BetterFirework Module

IceTank:
This Module adds better Fireworks utility. Features include:

* Auto takeoff while standing or falling by interacting with fireworks
* Auto takeoff on middle click (pick block key)
* Auto firework use on middle click
* Silent firework use from inventory with middle click


src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,187 @@
...
+    tag = ModuleTag.MOVEMENT,
+) {
+    private var fireworkInteract by setting("Firework Interact", true, "Automatically start flying when right clicking fireworks")
+    private var fireworkInteractCancel by setting("Firework Interact Cancel", false, "Cancel block interactions while holding fireworks", visibility = { fireworkInteract })
+    private var middleClick by setting("Middle Click", true, "Use firework on middle mouse click")
+    private var middleClickCancel by setting("Middle Click Cancel", false, description = "Cancel pick block action on middle mouse click", visibility = { middleClick })
+    private var clientSwing by setting("Swing", true, "Swing hand client side")
+    private var silentUse by setting("Silent", true, "Silent use fireworks from the inventory", visibility = { middleClick })
+
+    private var openDelay = 0
```
emyfops: Ideally you should never use raw values for durations. We have a Timer class to keep track of passed time and I had plans in the past to add ticks as a measurement unit

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,187 @@
...
+        if (canOpenElytra(player)) {
+            connection.sendPacket(ClientCommandC2SPacket(player, ClientCommandC2SPacket.Mode.START_FALL_FLYING))
+            startFirework(silentUse)
+            return true
+        }
+
+        return false
+    }
+
+    fun canOpenElytra(player: ClientPlayerEntity): Boolean {
```
emyfops: This can be replaced by a getter extending the ClientPlayerEntity

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    private fun SafeContext.getFireworkInInventoryScreen(): Int {
+        val screenHandler: PlayerScreenHandler = player.playerScreenHandler
+        for (i in PlayerScreenHandler.INVENTORY_START..<PlayerScreenHandler.INVENTORY_END) {
+            val itemStack = screenHandler.getSlot(i).stack
+            if (itemStack.item !== Items.FIREWORK_ROCKET) continue
+            return i
+        }
+        return -1
+    }
```
emyfops: We have utils to find and swap items

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    private var openDelay = 0
+
+    init {
+        listen<TickEvent.Pre> {
+            // Try opening the elytra after a delay after jumping
+            if (openDelay > 0) {
+                openDelay--
+                if (openDelay == 0) {
+                    tryTakeoff()
```
emyfops: I'm not sure about having all the logic being contained in one function and that function calling other functions. Unless you're a machine it's hard to keep track of the stack

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,187 @@
...
+import net.minecraft.network.packet.c2s.play.ClientCommandC2SPacket
+import net.minecraft.network.packet.c2s.play.HandSwingC2SPacket
+import net.minecraft.network.packet.c2s.play.PlayerInteractItemC2SPacket
+import net.minecraft.network.packet.c2s.play.UpdateSelectedSlotC2SPacket
+import net.minecraft.screen.PlayerScreenHandler
+import net.minecraft.screen.slot.SlotActionType
+import net.minecraft.util.Hand
+import net.minecraft.util.hit.HitResult
+
+object BetterFirework : Module(
```
emyfops, line 48: Wouldn't it be better to add this feature in the elytra flight module instead ?

emyfops (changes requested):

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
IceTank: This module only changes firework usage not elytra behavior

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
IceTank: I could not find any utils that tell me what inventory index the found items are in. We looked at this like 3 weeks ago. You couldn't find any either.

IceTank:
@Edouard127 so we merge?

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,190 @@
...
+    name = "BetterFirework",
+    description = "Automatic takeoff with fireworks",
+    tag = ModuleTag.MOVEMENT,
+) {
+    private var fireworkInteract by setting("Firework Interact", true, "Automatically start flying when right clicking fireworks")
+    private var fireworkInteractCancel by setting("Firework Interact Cancel", false, "Cancel block interactions while holding fireworks") { fireworkInteract }
```
beanbag44: Its annoying sometimes when you fly into a wall and try to use a firework for it to spam it on a block. Maybe this could be a general setting rather than an extension of fireworkInteract

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,190 @@
...
+import net.minecraft.screen.slot.SlotActionType
+import net.minecraft.util.Hand
+import net.minecraft.util.hit.HitResult
+
+object BetterFirework : Module(
+    name = "BetterFirework",
+    description = "Automatic takeoff with fireworks",
+    tag = ModuleTag.MOVEMENT,
+) {
+    private var fireworkInteract by setting("Firework Interact", true, "Automatically start flying when right clicking fireworks")
```
beanbag44: I would have used the name "Right Click Fly" instead of "Firework Interact"

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,190 @@
...
+        if (!silent) return false
+
+        stack.bestItemMatch(player.storage)
+            ?.let {
+                val swap = player.storage.indexOf(it)
+
+                interaction.clickSlot(player.playerScreenHandler.syncId, swap, 0, SlotActionType.SWAP, mc.player)
+
+                if (player.getInventory().selectedSlot != 0) {
+                    player.networkHandler.sendPacket(UpdateSelectedSlotC2SPacket(0))
```
beanbag44: this should request the hotbar manager with keep ticks set to 0 for silent swap

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,190 @@
...
+                return true
+            }
+
+        return false
+    }
+
+    enum class TakeoffState {
+        NONE,
+        JUMPING,
+        START_FLYING
```
beanbag44: these should be in camel case

beanbag44 (changes requested):
just small things. Also, for single line if blocks, its often cleaner to omit the braces and keep them on the same line or at least the action after the check. Like this:

if (something) doSomething()
else if (something else) doSomethingElse()

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
IceTank: You can just enable the Module

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,196 @@
...
+import net.minecraft.util.Hand
+import net.minecraft.util.hit.HitResult
+
+object BetterFirework : Module(
+    name = "BetterFirework",
+    description = "Automatic takeoff with fireworks",
+    tag = ModuleTag.MOVEMENT,
+) {
+    private var fireworkInteract by setting("Right Click Fly", true, "Automatically start flying when right clicking fireworks").group(Group.General)
+    private var fireworkInteractCancel by setting("Right Click Cancel", false, "Cancel block interactions while holding fireworks") { fireworkInteract }.group(Group.General)
```
beanbag44: i meant that this shouldn't be dependent on right click fly to be a feature. This should be its own option, separate from firework interact

beanbag44 (changes requested):
just one thing, aside from that i think its good

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    /**
+     * Returns true if the mc item interaction should be canceled
+     */
+    @JvmStatic
+    fun onInteract() =
+        runSafe {
+            if (!fireworkInteract) return false
```
beanbag44: a big when statement could be used here

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,247 @@
...
+            return cancelInteract
+        } ?: false
+
+    /**
+     * Returns true when the pick interaction should be canceled.
+     */
+    @JvmStatic
+    fun onPick() =
+        runSafe {
+            if (mc.crosshairTarget?.type == HitResult.Type.BLOCK && !middleClickCancel) {
```
beanbag44: again, could use a when statement

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -0,0 +1,247 @@
...
+
+        stack.bestItemMatch(player.hotbarAndStorage)
+            ?.let {
+                val swapSlotId = player.hotbarAndStorage.indexOf(it)
+                val hotbarSlotToSwapWith = player.hotbar.find { slot -> slot.isEmpty } ?.let { slot -> player.hotbar.indexOf(slot) } ?: 8
+
+                inventoryRequest {
+                    swap(swapSlotId, hotbarSlotToSwapWith)
+                    action {
+                        HotbarManager.request(HotbarRequest(hotbarSlotToSwapWith, this@BetterFirework, keepTicks = 0, nowOrNothing = true))
```
beanbag44: .done.let {} will run regardless of the value of done. Also, its probably cleaner to do HotbarRequest().submit()

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
beanbag44: i'd just save the request to a local variable, and check if (request.done) {}

beanbag44 (approved):

IceTank:
@emyfops you approve?

emyfops (approved):
lgtm

PR #158, opened 2025-10-13
Improvement: Editable config groups

beanbag44:
Provides a class that all config groups extend in order to allow editing the settings within. This allows modules like scaffold to change existing settings' names, descriptions, visibility, default value, etc, to better fit the use of the module.


emyfops (approved):
image

Wait until I can try it myself and then we can merge

PR #165, opened 2025-11-15
Fix coordinate hud element in the most scuffed way possible

IceTank:
pls merge

### DIFF
--- a/src/main/kotlin/com/lambda/config/groups/FormatterConfig.kt
+++ b/src/main/kotlin/com/lambda/config/groups/FormatterConfig.kt
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2025 Lambda
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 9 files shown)


emyfops (commented):
I will be adding a formatter settings group soon

emyfops:
I am gonna rework this a bit, I am not entirely satisfied as it only cover a single use

IceTank:
Could add a utility method to convert any coordinates to text. Could also be an easy way to obfuscate coordinates if everyone uses the utility method. This could prevent some people (@beanbag44) to accidentally printer coordinates in chat every once in a while (every few seconds). That would decrease the chance of leaking your own coords (you correctly leak your coords constantly).
Some people have custom modules that auto hide coordinates when you teleport a certain distance. We can add something similar that coord obfuscation will automatically turn on when you get teleported above a certain distance.

IceTank:
@emyfops #167

PR #173, opened 2025-12-07
Add ElytraAttitudeControl module

IceTank:
Adds Attitude control for elytra flying

Key features:
* Fly at a specified altitude by controlling pitch
* Fly at a specified speed by controlling pitch
* Automatically use fireworks when speed or altitude drops
* Features a PID controller for a smooth generalized pitch control
* Has a pitch40 mode to gain height so you can fly indefinitely


IceTank:
~~Still have to work a bit on the firework usage. I don't think it works right now.~~
Works now

PR #174, opened 2025-12-07
[Draft] Pitch40: Add Pitch40

IceTank:
I think ill integrate this into ElytraAlttitudeControl so don't merge yet


IceTank:
I'll be adding pitch40 to #173 instead

PR #187, opened 2025-12-13
improvement(AutoTotem): move cursor item before swap

cattyngmd:


emyfops:
Can you explain why this change is helpful ?

cattyngmd:
When you have an item in ur cursor you cant perform the offhand swap. So it will put down the item bcs otherwise you will die while holding it

PR #192, opened 2025-12-15
feat: add more hud settings

cattyngmd:
added following setting to hud modules:
- `Show Keybind` to `ModuleList`
- `Average` to `FPS`


beanbag44:
Looks good, but sameLine() still gets called even when showKeybind is disabled, resulting in all modules being placed on the same line

beanbag44 (changes requested):
looks good, just the formatting for sameLine seems off. Thats all though

cattyngmd:
wtf it looked fine on my phone, never using github there

beanbag44 (approved):
👍

emyfops:
Could you use a LimitedDecayQueue for the average fps count? Could you also use monotonic kotlin times instead of longs?

cattyngmd:
LimitedDecayQueue uses millis so theres no point of using it in that case

emyfops:
> LimitedDecayQueue uses millis so theres no point of using it in that case

I meant to use the queue to store nanos, either way it shouldnt really matter for the accuracy

PR #193, opened 2025-12-15
fix(PlayerTrap): account epsilon when calculating player hitbox size

cattyngmd:


cattyngmd:
also i think its better to move the `withEpsilon` method to some util for future use

beanbag44:
Thanks, I tried this before, but it seemed to break something. However, this seems to work. withEpsilon could be moved to util/extension/Box and maybe rename it to shrinkByEpsilon or something similar. Prob also best to inline the function directly to shrinkByEpsilon(): Box = expand(-1e-7) as it's cleaner.

beanbag44 (approved):
looks good!

emyfops:
@cattyngmd why do we need to shrink the box by the smallest minecraft unit ? (or whatever this is called)

cattyngmd:
bcs minecraft shrinks entity collision box by epsilon when checking is it possible to place block. without accounting that, the surround will extend itself when it shouldn't
изображение

PR #199, opened 2025-12-20
Feat: Added prefix command.

KMatias123:
Prefix command, `;prefix <something>`

People expect it and every other client has it :3

### DIFF
--- a/src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
+++ b/src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2025 Lambda
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.lambda.command.commands
+
+import com.lambda.brigadier.CommandResult.Companion.failure
+import com.lambda.brigadier.CommandResult.Companion.success
+import com.lambda.brigadier.argument.string
+import com.lambda.brigadier.argument.value
+import com.lambda.brigadier.execute
+import com.lambda.brigadier.executeWithResult
+import com.lambda.brigadier.required
+import com.lambda.command.CommandRegistry
+import com.lambda.command.LambdaCommand
+import com.lambda.config.Configuration
+import com.lambda.config.Setting
+import com.lambda.util.Communication.info
+import com.lambda.util.extension.CommandBuilder
+import com.lambda.util.text.buildText
+import com.lambda.util.text.literal
+
+object PrefixCommand : LambdaCommand("prefix",
+	usage = "prefix <prefix>",
+	description = "Sets the prefix for Lambda commands. If the prefix does not seem to work, try putting it in double quotes."
+) {
+	// i have no idea why someone would want to use some of these as a prefix
+	// but ig the people who run 20 clients at once could benefit from this
+	val ptrn = Regex("^[!\"#$%&'()*+,\\-./:;<=>?@\\[\\\\\\]^_`{|}~]$")
+	override fun CommandBuilder.create() {
+		required(string("prefix")) { prefixStr ->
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
```diff
@@ -0,0 +1,66 @@
...
+				val prefixChar = prefix.first()
+				val configurable = Configuration.configurableByName("command")?: return@executeWithResult failure("No command configurable found.")
+				val setting = (configurable.settings.find { it.name == "prefix" }?: return@executeWithResult failure("No prefix setting found.")) as Setting<*, Char>
+				setting.trySetValue(prefixChar)
+				return@executeWithResult success()
+			}
+		}
+
+		execute {
+			this@PrefixCommand.info(
```
emyfops: You don't need to specify the receiver for `info` as it is an extension of `Any`

src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
```diff
@@ -0,0 +1,66 @@
...
+	override fun CommandBuilder.create() {
+		required(string("prefix")) { prefixStr ->
+			executeWithResult {
+				val prefix = prefixStr().value()
+				if (!ptrn.matches(prefix)) {
+					return@executeWithResult failure("Prefix must be a single non-alphanumeric ASCII character, excluding spaces.")
+				}
+				val prefixChar = prefix.first()
+				val configurable = Configuration.configurableByName("command")?: return@executeWithResult failure("No command configurable found.")
+				val setting = (configurable.settings.find { it.name == "prefix" }?: return@executeWithResult failure("No prefix setting found.")) as Setting<*, Char>
```
emyfops: You can safe cast using `as?` instead of wrapping the setting with parenthesis and returning if it is null. Instead you safe cast and returns if the cast is also null

src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
```diff
@@ -0,0 +1,66 @@
...
+	val ptrn = Regex("^[!\"#$%&'()*+,\\-./:;<=>?@\\[\\\\\\]^_`{|}~]$")
+	override fun CommandBuilder.create() {
+		required(string("prefix")) { prefixStr ->
+			executeWithResult {
+				val prefix = prefixStr().value()
+				if (!ptrn.matches(prefix)) {
+					return@executeWithResult failure("Prefix must be a single non-alphanumeric ASCII character, excluding spaces.")
+				}
+				val prefixChar = prefix.first()
+				val configurable = Configuration.configurableByName("command")?: return@executeWithResult failure("No command configurable found.")
```
emyfops: I won't let this pr in if you don't put a space between the question mark and the parenthesis :3333

src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.lambda.util.text.buildText
+import com.lambda.util.text.literal
+
+object PrefixCommand : LambdaCommand("prefix",
+	usage = "prefix <prefix>",
+	description = "Sets the prefix for Lambda commands. If the prefix does not seem to work, try putting it in double quotes."
+) {
+	// i have no idea why someone would want to use some of these as a prefix
+	// but ig the people who run 20 clients at once could benefit from this
+	val ptrn = Regex("^[!\"#$%&'()*+,\\-./:;<=>?@\\[\\\\\\]^_`{|}~]$")
```
emyfops, line 44: Can you add a new line after the val ?

emyfops (changes requested):

src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
KMatias123: like
```
val
ptrn = Regex("^[!\"#$%&'()*+,\\-./:;<=>?@\\[\\\\\\]^_`{|}~]$")
```
? is that even valid syntax?
e: oh nvm i get what you mean i am tired :3

src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
KMatias123: space added

src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
KMatias123: removed

src/main/kotlin/com/lambda/command/commands/PrefixCommand.kt
KMatias123: it is correct now right?

emyfops (approved):

emyfops:
@KMatias123 change the `prefix` to `greedyString` and this will be good to merge

emyfops (approved):

PR #200, opened 2025-12-20
Feat: KillAura DPS calculation

KMatias123:
I don't want the KillAura to always switch to the axe if I have swap on, so this fixes that.
It prefers the axe or the tool with the highest damage/hit if it is in the "Hit Damage" mode if you like the old behavior.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/combat/KillAura.kt
+++ b/src/main/kotlin/com/lambda/module/modules/combat/KillAura.kt
@@ -34,6 +34,7 @@ import com.lambda.task.RootTask.run
 import com.lambda.threading.runSafeAutomated
 import com.lambda.util.NamedEnum
 import com.lambda.util.item.ItemStackUtils.attackDamage
+import com.lambda.util.item.ItemStackUtils.attackSpeed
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/combat/KillAura.kt
```diff
@@ -76,6 +78,12 @@ object KillAura : Module(
         Delay
     }
 
+    @Suppress("unused")
+    enum class DamageMode(override val displayName: String) : NamedEnum {
+        DPS("Damage Per Second"),
+        Total("Hit Damage")
```
emyfops: You could inline the check in the code inside the enum. Ex:
```
enum class DamageMode(override val displayName: String, val block: SafeContext.(ItemStack) -> Double) : NamedEnum {
DPS("Damage Per Second", { player.attackDamage(stack = it) * player.attackSpeed(stack = it) }),
}
```

to then simply sort with the block

emyfops (changes requested):

emyfops (approved):

Issue #203, opened 2025-12-21
Bug: Friend manager serialization issue


cattyngmd:
### What happened?

the client didnt save the friend list correctly
<img width="329" height="215" alt="Image" src="https://github.com/user-attachments/assets/af568f0c-f4c4-45ce-b69d-8f2b746a4660" />

### What should have happened?

i can assume the friend list file should have a model like this
```json

{
  "friends": {
    "friends": [
       {
         "name": "string",
         "uuid": "string"
      }
    ]
  }
}
```

### Version

1.21.5

emyfops:
Should be fixed in #204 can you test ?

cattyngmd:
@emyfops yep it works now

PR #204, opened 2025-12-21
Fix: CollectionSettings mapping stringified values instead of converting the collection to a tree

emyfops:
This pull request modifies slightly comparable collection settings by converting the entire tree directly instead of mapping the collection values to their stringified representations. Collection setting of comparable types was removed and a new flag was introduce to chose the setting core before creating the setting. This pull request also fixes the friend list being serialized using the class collection setting core.


emyfops:
closes #203

PR #205, opened 2025-12-21
Fix: Tracer origin going offscreen

KMatias123:
If you try to sprint with view bobbing on and tracers in the crosshair mode the center where they all "point" will move around due to view bobbing. This is annoying. This commit fixes it in a way that just moves the "start of the tracers" away from the camera so the movement is not as noticeable. In the future it can also be fixed by completely negating the movement from the `camera.horizontalPlane`.
This is the kotlin code for calculating the bobbing if someone wants to calculate the bobbing and implement a better solution.

```kotlin
val clientPlayerLikeState: ClientPlayerLikeState = player.state
val reverseSpeed = clientPlayerLikeState.getReverseLerpedDistanceMoved(tickDelta)
val progress = clientPlayerLikeState.lerpMovement(tickDelta)

val verticalBob = abs(cos(reverseSpeed * PI.toFloat() - 0.2f) * progress) * 0.1f
val horizontalBob = sin(reverseSpeed * PI.toFloat()) * progress * 0.05f

val result = Vec3d(horizontalBob.toDouble(), verticalBob.toDouble(), -1.0)
```
I just put this here cuz I don't want anyone else to have to be without tracers :3

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/render/EntityESP.kt
+++ b/src/main/kotlin/com/lambda/module/modules/render/EntityESP.kt
@@ -17,7 +17,6 @@
 
 package com.lambda.module.modules.render
 
-import com.lambda.Lambda.mc
 import com.lambda.context.SafeContext
 import com.lambda.event.events.GuiEvent
 import com.lambda.event.events.RenderEvent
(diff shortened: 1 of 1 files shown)


emyfops:
This would be a temporary fix as constructor plans to add 2d tracers

emyfops:
This pr is probably superseded by #230. I'll close it and you can open it back in case this is still an issue.

PR #208, opened 2025-12-22
Fix: Transparency for block colors in blockesp

KMatias123:
Now it has transparency when using the block color mode. And custom colors for portals and Ender Chests cuz they don't have good colors from the map color map.

### DIFF
--- a/src/main/kotlin/com/lambda/util/extension/World.kt
+++ b/src/main/kotlin/com/lambda/util/extension/World.kt
@@ -53,8 +53,15 @@ fun SafeContext.collisionShape(state: BlockState, pos: BlockPos): VoxelShape =
 fun SafeContext.outlineShape(state: BlockState, pos: BlockPos) =
     state.getOutlineShape(world, pos).offset(pos)
 
-fun SafeContext.blockColor(state: BlockState, pos: BlockPos) =
-    Color(state.getMapColor(world, pos).color)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/lambda/util/extension/World.kt
```diff
@@ -53,8 +53,15 @@ fun SafeContext.collisionShape(state: BlockState, pos: BlockPos): VoxelShape =
 fun SafeContext.outlineShape(state: BlockState, pos: BlockPos) =
     state.getOutlineShape(world, pos).offset(pos)
 
-fun SafeContext.blockColor(state: BlockState, pos: BlockPos) =
-    Color(state.getMapColor(world, pos).color)
+fun SafeContext.blockColor(state: BlockState, pos: BlockPos): Color {
+    return when (state.block) {
+        Blocks.ENDER_CHEST -> Color(0xFF00FF)
+        Blocks.NETHER_PORTAL -> Color(0xaa00aa)
+        Blocks.END_PORTAL -> Color(0xFF00FF)
```
cattyngmd, line 60: i think pink is not the right color for end portal

KMatias123: It looks "darker" when the alpha is set to something, if you want it to be something entirely different do you want to suggest something?

PR #215, opened 2026-01-02
Feat: ChatTimestamp, CustomChat, FancyChat, FriendHighlight

emyfops:


beanbag44 (approved):
lgtm

beanbag44:
although i'd like to change the default fancy chat from "uwu"

emyfops:
> although i'd like to change the default fancy chat from "uwu"

the "uwu" shall stay uwu. we need more catgirls and puppygirls on 2b2t.

PR #217, opened 2026-01-06
Feat: Client module toggles for loading screen and UI sounds

ignchaadi:
I have introduced a new Client module (in the Client category) that acts as a home for toggles. The module itself has no standalone functionality; it exists solely to expose settings that gate existing client behavior. Currently, it includes a toggle to enable/disable Lambda's custom splash screen, as well as a toggle to enable/disable Lambda's UI sounds.

Due to how the splash overlay and resource reload lifecycle work, the vanilla splash screen may only be fully visible on the subsequent reload after disabling the Lambda splash screen (e.g. a second F3+T). Enabling the Lambda splash screen applies immediately.

### DIFF
--- a/src/main/java/com/lambda/mixin/render/SplashOverlayMixin.java
+++ b/src/main/java/com/lambda/mixin/render/SplashOverlayMixin.java
@@ -17,6 +17,7 @@
 
 package com.lambda.mixin.render;
 
+import com.lambda.module.modules.client.Client;
 import com.lambda.util.LambdaResourceKt;
 import net.minecraft.client.gui.screen.SplashOverlay;
 import net.minecraft.client.texture.ReloadableTexture;
@@ -25,10 +26,7 @@ import net.minecraft.util.Identifier;
 import net.minecraft.util.math.ColorHelper;
 import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
 import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
-import org.spongepowered.asm.mixin.Final;
-import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Mutable;
-import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.*;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/lambda/sound/SoundManager.kt
```diff
@@ -34,6 +35,7 @@ object SoundManager : Loadable {
     }
 
     fun playSoundRandomly(event: SoundEvent, pitch: Double = 1.0, pitchRange: Double = 0.05) {
+        if (!Client.toggleSounds) return
```
emyfops: This is not the ideal solution. Under no circumstances should a module dictate the behavior if an unrelated function. I recommend we check for the condition in the `onEnable` and `onDisable` in the `Module` class

src/main/java/com/lambda/mixin/render/SplashOverlayMixin.java
```diff
@@ -61,7 +67,12 @@ public LogoTextureMixin(Identifier location) {
 
         @WrapOperation(method = "loadContents", at = @At(value = "INVOKE", target = "Lnet/minecraft/resource/ResourceFactory;open(Lnet/minecraft/util/Identifier;)Ljava/io/InputStream;"))
         InputStream wrapLoadTextureData(ResourceFactory instance, Identifier id, Operation<InputStream> original) {
-            return LambdaResourceKt.getStream("textures/lambda_banner.png");
+            if ("lambda".equals(id.getNamespace())
+                    && "textures/lambda_banner.png".equals(id.getPath())) {
```
emyfops, line 71: I'm not sure to understand this condition here. Could you explain it ?

emyfops (changes requested):

src/main/kotlin/com/lambda/sound/SoundManager.kt
ignchaadi: This is reasonable, I unfortunately was on a very tight timeframe when adding these and figured it was the easiest solution at the time. I can move the check into the `Module` class and resubmit once I find the time.

src/main/java/com/lambda/mixin/render/SplashOverlayMixin.java
ignchaadi: During my testing, resource reload would replace the Lambda banner with a blue box rather than the vanilla banner, which was persistent across resource reloads. That check fixed it for reasons I'm unsure of, as again I was on a tight timeframe and could not look into it further. Could definitely be a local issue, I'll look into it.

ignchaadi:
Forgot to change commit message, moved the check into `Module.kt` in `onEnable` and `onDisable`, as well as their respective "Unsafe" variants.

beanbag44:
disabling sounds doesn't seem to disable the click gui open and close sound

beanbag44:
also i'm not sure if this is just me but if i disable the custom screen, it shows for half the time, then swaps to the mojang image

ignchaadi:
> disabling sounds doesn't seem to disable the click gui open and close sound

This is due to the checks being moved into `Module.kt`. If the functionality is desired, I can push changes tomorrow.

ignchaadi:
> also i'm not sure if this is just me but if i disable the custom screen, it shows for half the time, then swaps to the mojang image

This is due to the way resource reload works, I briefly looked into it and was able to partially fix it (it was a blue box previously). This is something I did not have time to look into further. It only occurs on the first reload, and sometimes on the second. Subsequent reloads do not experience this bug. I can look into it more this weekend, as I will have more free time. 

If desired, I can hold off on pushing the changes mentioned previously until I fix the bug, and push both simultaneously.

PR #220, opened 2026-01-11
Feature: Mutable module priority 

beanbag44:
Makes the priority field in listeners an updatableLazy so that modules can adjust their priority and update the sorting within the listener collections.

A modules priority, if being set (which is only necessary if it might conflict with other modules), should be set at the beginning of the init block. This is just as a default and can be changed by the user in-game.

The listener collections are updated by removing the owners listeners, updating their priority, and adding them back again. This minimizes cost, and resorting only occurs when the user changes a module priority


emyfops (changes requested):
We will need to discuss to find another solution. This feels extremely hacky and does not seem fine for the long run. We can't just add hacks on top of the existing event system. I hate to be that person but I would much rather ship features at low rate than merging lots of hacks.

beanbag44:
We could change the direct Module reference to an interface that Module implements, which could then also automatically detect priority change and internally update the sorting. I think to call this extremely hacky is a little excessive. In the end, i think the only thing i truly dislike about the changes are that updating the listener sorting is down to whoever changes the priority. Once again though, this can be fixed with an interface. If its the actual method of updating the sorting you dont like, perhaps we could look into a new collection type which could sort on demand? I dont like adding hacky features either, and i dont think this would be one, maybe just some changes to improve its implementation.

PR #227, opened 2026-01-26
FIX: Prevents betterfirework from trying to start flying without an elytra or a firework. Shows a warning message to the user instead.

KMatias123:
Previously BetterFirework tried to fly without an Elytra. This fixes the issue.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
@@ -33,7 +33,6 @@ import com.lambda.threading.runSafe
 import com.lambda.util.Communication.warn
 import com.lambda.util.KeyCode
 import com.lambda.util.Mouse
-import com.lambda.util.item.ItemStackUtils.copy
 import com.lambda.util.player.SlotUtils.hotbarAndInventoryStacks
 import com.lambda.util.player.SlotUtils.hotbarStacks
 import net.minecraft.client.network.ClientPlayerEntity
@@ -52,18 +51,18 @@ object BetterFirework : Module(
 ) {
 	private var activateButton by setting("Activate Key", Bind(0, 0, Mouse.Middle.ordinal), "Button to activate Firework")
 		.onPress {
-			mc.player?.isElytraEquipped?.let { equipped ->
+			player.isElytraEquipped.let { equipped ->
 				if (!equipped) {
 					warn("You need to equip an elytra to use this module!")
 					return@onPress
 				}
-			} ?: return@onPress
-			mc.player?.hasFireworks?.let { hasFireworks ->
+			}
+			player.hasFireworks.let { hasFireworks ->
 				if (!hasFireworks) {
 					warn("You need to have fireworks in your inventory to use this module!")
 					return@onPress
 				}
-			} ?: return@onPress
+			}
 			// Prevent using multiple times
 			if (takeoffState != TakeoffState.None) return@onPress
 			// If already gliding use another firework
--- a/src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
@@ -33,6 +33,7 @@ import com.lambda.threading.runSafe
 import com.lambda.util.Communication.warn
 import com.lambda.util.KeyCode
 import com.lambda.util.Mouse
+import com.lambda.util.item.ItemStackUtils.copy
 import com.lambda.util.player.SlotUtils.hotbarAndInventoryStacks
 import com.lambda.util.player.SlotUtils.hotbarStacks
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -43,35 +45,64 @@ import net.minecraft.util.Hand
...
-    val ClientPlayerEntity.canTakeoff: Boolean
-        get() = isOnGround || canOpenElytra
-
-    val ClientPlayerEntity.canOpenElytra: Boolean
-        get() = !abilities.flying && !isClimbing && !isGliding && !isTouchingWater && !isOnGround && !hasVehicle() && !hasStatusEffect(StatusEffects.LEVITATION)
-
-    init {
+	private var activateButton by setting("Activate Key", Bind(0, 0, Mouse.Middle.ordinal), "Button to activate Firework")
+		.onPress {
+			mc.player?.isElytraEquipped()?.let { equipped ->
```
emyfops: You don't have to use `mc` as `player` is part of the safe context.

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -43,35 +45,64 @@ import net.minecraft.util.Hand
...
+	private var invUse by setting("Inventory", true, "Use fireworks from inventory") { activateButton.key != KeyCode.Unbound.code }
+
+	private var takeoffState = TakeoffState.None
+
+	fun ClientPlayerEntity.isElytraEquipped(): Boolean {
+		this.inventory?.equipment?.get(EquipmentSlot.CHEST)?.let { return it.item == Items.ELYTRA }
+		return false
+	}
+
+	fun ClientPlayerEntity.hasFireworks(minimum: Int = 1): Boolean {
```
emyfops: You can instead use this

```kt
    val ClientPlayerEntity.hasFireworks: Boolean
        get() = selectStack { isItem(Items.FIREWORK_ROCKET) }
            .filterStacks(inventory.mainStacks)
            .isNotEmpty()
```

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -43,35 +45,64 @@ import net.minecraft.util.Hand
...
+	private var middleClickCancel by setting("Middle Click Cancel", false, description = "Cancel pick block action on middle mouse click") { activateButton.key != KeyCode.Unbound.code }
+	private var fireworkInteract by setting("Right Click Fly", true, "Automatically start flying when right clicking fireworks")
+	private var fireworkInteractCancel by setting("Right Click Cancel", false, "Cancel block interactions while holding fireworks") { fireworkInteract }
+
+	private var clientSwing by setting("Swing", true, "Swing hand client side")
+	private var invUse by setting("Inventory", true, "Use fireworks from inventory") { activateButton.key != KeyCode.Unbound.code }
+
+	private var takeoffState = TakeoffState.None
+
+	fun ClientPlayerEntity.isElytraEquipped(): Boolean {
```
emyfops: We prefer using getters instead of functions that only returns a value
Here's how that would look like:
```kt
    val ClientPlayerEntity.hasElytraEquipped: Boolean
        get() = inventory.equipment.get(EquipmentSlot.CHEST).item == Items.ELYTRA
```

emyfops (changes requested):

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
KMatias123: I was thinking that this could be useful in the future development of the module. That is why I added the minimum required fireworks, this could be moved to StackSelection.kt with an Item argument though. What do you think?

emyfops: > I was thinking that this could be useful in the future development of the module. That is why I added the minimum required fireworks, this could be moved to StackSelection.kt with an Item argument though. What do you think?

There is already an argument in `selectStack` that specifies the minimum stack size. It is currently set to at least 1

src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -43,35 +45,56 @@ import net.minecraft.util.Hand
...
-    init {
+	private var activateButton by setting("Activate Key", Bind(0, 0, Mouse.Middle.ordinal), "Button to activate Firework")
+		.onPress {
+			player.isElytraEquipped.let { equipped ->
+				if (!equipped) {
+					warn("You need to equip an elytra to use this module!")
+					return@onPress
+				}
+			}
+			player.hasFireworks.let { hasFireworks ->
```
emyfops: You can inline those checks now that you're not checking for null :3

emyfops (approved):

PR #228, opened 2026-01-26
Ref: Mixin cleanup

emyfops:
~~Due to the nature of `Inject`, they can cause mixin conflicts with mods that aren't up to date or don't use mixinextra features. This pull request replaces *most* injections by wrapping operations~~

~~This pull request needs extensive testing as it may break current features and introduce logic bugs~~

This pull request cleans up the mixins and replaces some mixins with less intrusive ones


0-x-2-2:
I don't know why you would use this over `@inject` unless you are doing some weird `@redirect` style injection.

src/main/java/com/lambda/mixin/baritone/BaritonePlayerContextMixin.java
```diff
@@ -36,13 +35,11 @@ public class BaritonePlayerContextMixin {
...
-    @Inject(method = "playerRotations", at = @At("HEAD"), cancellable = true, remap = false)
-    void syncRotationWithBaritone(CallbackInfoReturnable<Rotation> cir) {
-        if (baritone != BaritoneManager.getPrimary()) return;
+    @WrapMethod(method = "playerRotations", remap = false)
+    Rotation syncRotationWithBaritone(Operation<Rotation> original) {
+        if (baritone != BaritoneManager.getPrimary())
+            return original.call();
 
-        RotationManager rm = RotationManager.INSTANCE;
-        cir.setReturnValue(new Rotation(
```
0-x-2-2, line 44: This may be an example of where you should use `@ModifyReturnValue(method = "playerRotations", at = @At("RETURN"))`, not `@WrapMethod`, as long as there are no side effects from the original function.

src/main/java/com/lambda/mixin/baritone/LookBehaviourMixin.java
```diff
@@ -23,37 +23,34 @@
...
-    @Inject(method = "updateTarget", at = @At("HEAD"), cancellable = true)
-    void onTargetUpdate(Rotation rotation, boolean blockInteract, CallbackInfo ci) {
-        if (instance.baritone != BaritoneManager.getPrimary()) return;
+    @WrapMethod(method = "updateTarget")
+    void onTargetUpdate(Rotation rotation, boolean blockInteract, Operation<Void> original) {
+        if (instance.baritone != BaritoneManager.getPrimary())
+            original.call(rotation, blockInteract);
 
         RotationManager.handleBaritoneRotation(rotation.getYaw(), rotation.getPitch());
-        ci.cancel();
```
0-x-2-2, line 43: Effectively, you are doing an overwrite anyway. Generally, the use case of `@WrapMethod` is to wrap a method with locks or try catch or running code before and after, even if someone is cancelling early. Most of the time, `@Inject` is likely the correct solution.

src/main/java/com/lambda/mixin/baritone/LookBehaviourMixin.java
```diff
@@ -23,37 +23,34 @@
...
     }
 
-    @Inject(method = "onPlayerUpdate", at = @At("HEAD"), cancellable = true)
-    void onUpdate(PlayerUpdateEvent event, CallbackInfo ci) {
-        if (instance.baritone != BaritoneManager.getPrimary()) return;
-
-        ci.cancel();
```
0-x-2-2, line 50: https://github.com/lambda-client/lambda/pull/228/changes#r2730807085

src/main/java/com/lambda/mixin/baritone/LookBehaviourMixin.java
```diff
@@ -23,37 +23,34 @@
...
+    void onUpdate(PlayerUpdateEvent event, Operation<Void> original) {
+        if (instance.baritone != BaritoneManager.getPrimary())
+            original.call(event);
     }
 
-    @Inject(method = "onPlayerRotationMove", at = @At("HEAD"), cancellable = true)
-    void onMovementUpdate(RotationMoveEvent event, CallbackInfo ci) {
-        if (instance.baritone != BaritoneManager.getPrimary()) return;
-
-        ci.cancel();
```
0-x-2-2, line 57: https://github.com/lambda-client/lambda/pull/228/changes#r2730807085

src/main/java/com/lambda/mixin/client/sound/SoundSystemMixin.java
```diff
@@ -19,20 +19,19 @@
...
-import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
-import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
 @Mixin(SoundSystem.class)
 public class SoundSystemMixin {
-    @Inject(method = "play(Lnet/minecraft/client/sound/SoundInstance;)Lnet/minecraft/client/sound/SoundSystem$PlayResult;", at = @At("HEAD"), cancellable = true)
-    public void onPlay(SoundInstance sound, CallbackInfoReturnable<SoundSystem.PlayResult> cir) {
-        if (EventFlow.post(new ClientEvent.Sound(sound)).isCanceled()) {
-            cir.setReturnValue(SoundSystem.PlayResult.NOT_STARTED);
```
0-x-2-2, line 35: This isn't any different from the original; it basically only prevents hooks from other mods from being called if they were injected at the head before you.

src/main/java/com/lambda/mixin/entity/ClientPlayerEntityMixin.java
```diff
@@ -140,13 +144,14 @@ boolean modifyIsSprinting(boolean original) {
         return EventFlow.post(new MovementEvent.Sprint(original)).getSprint();
     }
 
-    @Inject(method = "isSneaking", at = @At(value = "HEAD"), cancellable = true)
```
0-x-2-2, line 143: Use `@ModifyReturnValue` instead of `@WrapMethod` since this function has no side effects.

src/main/java/com/lambda/mixin/entity/ClientPlayerEntityMixin.java
```diff
@@ -129,8 +132,9 @@ private PlayerMoveC2SPacket.OnGroundOnly wrapOnGroundOnly(boolean onGround, bool
         return original.call(moveEvent.getOnGround(), moveEvent.isCollidingHorizontally());
     }
 
-    @Inject(method = "sendMovementPackets", at = @At("TAIL"))
-    private void injectSendMovementPacketsReturn(CallbackInfo ci) {
```
0-x-2-2, line 133: TAIL/RETURN `@Inject` injections are always safe unless you want to always run this if someone cancels this function with `@Inject` much earlier, which is likely not what you want.

src/main/java/com/lambda/mixin/entity/ClientPlayerEntityMixin.java
```diff
@@ -140,13 +144,14 @@ boolean modifyIsSprinting(boolean original) {
...
-    @Inject(method = "isSneaking", at = @At(value = "HEAD"), cancellable = true)
-    void injectSneakingInput(CallbackInfoReturnable<Boolean> cir) {
+    @WrapMethod(method = "isSneaking")
+    boolean injectSneakingInput(Operation<Boolean> original) {
         ClientPlayerEntity self = (ClientPlayerEntity) (Object) this;
-        if (self != Lambda.getMc().player) return;
 
-        if (self.input == null) return;
-        cir.setReturnValue(EventFlow.post(new MovementEvent.Sneak(self.input.playerInput.sneak())).getSneak());
+        if (self != Lambda.getMc().player) return false;
```
0-x-2-2: This seems incorrect.

src/main/java/com/lambda/mixin/entity/ClientPlayerEntityMixin.java
```diff
@@ -183,9 +188,10 @@ private void wrapSwingHand(ClientPlayerEntity instance, Hand hand, Operation<Voi
         viewModel.adjustSwing(hand, instance);
     }
 
-    @Inject(method = "updateHealth", at = @At("HEAD"))
-    public void injectUpdateHealth(float health, CallbackInfo ci) {
```
0-x-2-2, line 187: This was perfectly safe.

src/main/java/com/lambda/mixin/entity/ClientPlayInteractionManagerMixin.java
```diff
@@ -52,44 +51,55 @@ public class ClientPlayInteractionManagerMixin {
     @Shadow
     public int lastSelectedSlot;
 
-    @Inject(method = "interactBlock", at = @At("HEAD"), cancellable = true)
-    public void interactBlockHead(final ClientPlayerEntity player, final Hand hand, final BlockHitResult hitResult, final CallbackInfoReturnable<ActionResult> cir) {
-        if (EventFlow.post(new PlayerEvent.Interact.Block(hand, hitResult)).isCanceled()) {
-            cir.setReturnValue(ActionResult.FAIL);
-        }
+    @WrapMethod(method = "interactBlock")
```
0-x-2-2: I don't know why you did most of these; it just means this will always execute even if someone cancels.

src/main/java/com/lambda/mixin/entity/EntityMixin.java
```diff
@@ -160,18 +159,28 @@ private boolean wrapSetPitch(Entity instance, float yaw) {
         return RotationManager.getLockPitch() == null;
     }
 
-    @Inject(method = "isSprinting()Z", at = @At("HEAD"), cancellable = true)
-    private void injectIsSprinting(CallbackInfoReturnable<Boolean> cir) {
+    @WrapMethod(method = "isSprinting()Z")
+    private boolean injectIsSprinting(Operation<Boolean> original) {
```
0-x-2-2: @ModifyReturnValue

src/main/java/com/lambda/mixin/entity/EntityMixin.java
```diff
@@ -160,18 +159,28 @@ private boolean wrapSetPitch(Entity instance, float yaw) {
...
+        return original.call();
     }
 
-    @Inject(method = "getPose", at = @At("HEAD"), cancellable = true)
-    private void injectGetPose(CallbackInfoReturnable<EntityPose> cir) {
-        var entity = (Entity) (Object) this;
-        if (!(entity instanceof ClientPlayerEntity player)) return;
-        if (ElytraFly.INSTANCE.isEnabled() && ElytraFly.getMode() == ElytraFly.FlyMode.Bounce && player.isGliding()) cir.setReturnValue(EntityPose.GLIDING);
+    @WrapMethod(method = "getPose")
+    private EntityPose injectGetPose(Operation<EntityPose> original) {
```
0-x-2-2: @ModifyReturnValue

src/main/java/com/lambda/mixin/entity/HandledScreensMixin.java
```diff
@@ -18,21 +18,21 @@
...
-import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 @Mixin(HandledScreens.Provider.class)
 public interface HandledScreensMixin {
-    @Inject(method = "open", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/MinecraftClient;setScreen(Lnet/minecraft/client/gui/screen/Screen;)V"))
-    private void injectOpen(Text name, ScreenHandlerType<ScreenHandler> type, MinecraftClient client, int id, CallbackInfo ci) {
-        if (client.player == null) return;
-        InventoryManager.onSetScreenHandler(client.player.currentScreenHandler);
+    @WrapOperation(method = "open", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/MinecraftClient;setScreen(Lnet/minecraft/client/gui/screen/Screen;)V"))
```
0-x-2-2: I feel like this is wrong.

src/main/java/com/lambda/mixin/entity/LivingEntityMixin.java
```diff
@@ -62,11 +59,11 @@ public abstract class LivingEntityMixin extends EntityMixin {
      * }
      * }</pre>
      */
-    @Inject(method = "jump", at = @At("HEAD"), cancellable = true)
-    void onJump(CallbackInfo ci) {
+    @WrapMethod(method = "jump")
```
0-x-2-2: Very likely was fine before, like usual.

src/main/java/com/lambda/mixin/entity/LivingEntityMixin.java
```diff
@@ -85,18 +82,14 @@ void onJump(CallbackInfo ci) {
...
     }
 
-    @Inject(method = "travel", at = @At("HEAD"), cancellable = true)
-    void onTravelPre(Vec3d movementInput, CallbackInfo ci) {
-        if (EventFlow.post(new MovementEvent.Entity.Pre(lambda$instance, movementInput)).isCanceled()) {
-            ci.cancel();
+    @WrapMethod(method = "travel")
+    void onTravelPre(Vec3d movementInput, Operation<Void> original) {
+        if (!EventFlow.post(new MovementEvent.Entity.Pre(lambda$instance, movementInput)).isCanceled()) {
+            original.call(movementInput);
```
0-x-2-2: no reason to wrap this

0-x-2-2: Use @Inject or https://github.com/LlamaLad7/MixinExtras/wiki/ModifyReturnValue when you use the original result.

src/main/java/com/lambda/mixin/entity/LivingEntityMixin.java
```diff
@@ -185,9 +178,11 @@ private void wrapPushAwayFrom(Entity entity, Operation<Void> original) {
         original.call(entity);
     }
 
-    @Inject(method = "isGliding", at = @At("HEAD"), cancellable = true)
-    private void injectIsGliding(CallbackInfoReturnable<Boolean> cir) {
-        if (lambda$instance != Lambda.getMc().player) return;
-        cir.setReturnValue(ElytraFly.isGliding());
+    @WrapMethod(method = "isGliding")
+    private boolean injectIsGliding(Operation<Boolean> original) {
```
0-x-2-2: `@ModifyReturnValue`

0-x-2-2 (changes requested):
Most of these are wrong, and you shouldn't replace some of these. `@Inject` is perfectly safe and likely what you want *most* of the time.

0-x-2-2:
Most of this is just abusing the feature.

0-x-2-2:
But many of the old injections were also wrong and should have been using some other feature from mixinextras.

emyfops:
I mistakenly took an error from last year and attributed to `Inject` while the culprit was `Redirect`. I'll force remove my changes and instead use this pr to clean up some mixins.

emyfops:
I've tested for potential breaking changes and they're all good

PR #229, opened 2026-01-26
Feature InventoryCleaner module

IceTank:
This module adds a feature to auto drop items when you don't have space in your inventory


IceTank:
https://github.com/user-attachments/assets/9b15dd93-94f7-4b50-aed3-80a468e5aa59


https://github.com/user-attachments/assets/c1674892-f0c0-4943-92e0-fbe3cbb66860

src/main/java/com/lambda/mixin/render/HandledScreenMixin.java
```diff
@@ -36,6 +40,14 @@ private void onMouseClicked(Click click, boolean doubled, CallbackInfoReturnable
         }
     }
 
+    @Inject(method = "onMouseClick(Lnet/minecraft/screen/slot/Slot;IILnet/minecraft/screen/slot/SlotActionType;)V", at = @At(value = "HEAD"), cancellable = true)
```
beanbag44: prob best to replace this with an InventoryEvent for slot clicking

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
```diff
@@ -0,0 +1,121 @@
...
+    private val itemsCanTrash by setting("Items Can Trash", setOf<Item>(Items.NETHERRACK, Items.COBBLESTONE), ITEM.toSet())
+
+    private val dropToPickup by setting("Drop To Pickup", true)
+    private val itemsToPickup by setting("Items To Pickup", setOf(), ITEM.toSet())
+
+    private val timer = Timer()
+
+    init {
+        listen<TickEvent.Pre> {
+            if (!timer.timePassed(200.milliseconds)) {
```
beanbag44: we have an actions per second timer in the inventory config. I'd set the default value to this and let the inventory manager handle the delay

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
```diff
@@ -0,0 +1,121 @@
...
+                return@listen
+            }
+            checkShouldTrashSomething()
+        }
+    }
+
+    fun SafeContext.checkShouldTrashSomething() {
+        if (player.health > 0.0f && !player.isSpectator && dropToPickup && !player.isCreative) {
+            val box = if (player.vehicle != null && !player.vehicle!!.isRemoved) {
+                player.boundingBox.union(player.vehicle!!.boundingBox).expand(1.0, 0.0, 1.0)
```
beanbag44: we try to avoid using !! at all cost as it bypasses kotlins null safety and allows for NullPointerExceptions. Best to replace it with ?.let or set it to a variable which can be smart cast to not null if its immutable after your null check

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
```diff
@@ -0,0 +1,121 @@
...
+            } else {
+                player.boundingBox.expand(1.0, 0.5, 1.0)
+            }
+            val items = world.getEntitiesByType<ItemEntity>(EntityType.ITEM, box) { entity -> itemsToPickup.contains(entity.stack.item) && entity.isOnGround }.map { i -> i.stack.item }
```
beanbag44: the first lambda
```
entity -> itemsToPickup.....
```
can be split onto the next line
```
{ entity ->
    itemsToPickup... 
}.map {...
```

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
```diff
@@ -0,0 +1,121 @@
...
+            if (items.isNotEmpty() && InventoryContainer.spaceAvailable(StackSelection.selectStack { isOneOfItems(items) }) == 0) {
```
beanbag44: spaceAvailable can actually return -1 iirc so best to switch to <= 0 instead of == 0

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Called when a slot is clicked in a screen handler.
+     * Returns true if the click was handled
+     */
+    @JvmStatic
+    fun onClick(slotIndex: Int, button: Int, actionType: SlotActionType): Boolean =
+        runSafe {
+            if (!isEnabled) return false
+            if (actionType != SlotActionType.QUICK_MOVE || button != 0) return false
+            val screenHandler = player.currentScreenHandler
```
beanbag44: i dont mine the direct screen handler referencing too much here as our container utils are still lacking in some areas but perhaps we can pass the container in the slot click event i spoke about in the handled screen mixin comment.

beanbag44 (changes requested):
Really handy module, just some small nit picks and changes

src/main/java/com/lambda/mixin/render/HandledScreenMixin.java
emyfops: If I see ONE MORE inject I will tweak out

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
```diff
@@ -0,0 +1,121 @@
...
+import net.minecraft.entity.ItemEntity
+import net.minecraft.item.Item
+import net.minecraft.item.Items
+import net.minecraft.registry.Registries.ITEM
+import net.minecraft.screen.GenericContainerScreenHandler
+import net.minecraft.screen.ScreenHandler
+import net.minecraft.screen.slot.SlotActionType
+import kotlin.time.Duration.Companion.milliseconds
+
+object EasyTrash : Module(
```
emyfops: Don't forget to name the arguments

emyfops (changes requested):

src/main/java/com/lambda/mixin/render/HandledScreenMixin.java
IceTank: Replaced with @Overwrite :thumbs_up:

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
IceTank: What does that mean?

emyfops: Add named arguments

emyfops: Add named arguments

src/main/kotlin/com/lambda/module/modules/player/EasyTrash.kt
IceTank: How would I use the inventory request limit? Just spam inventory requests every tick?

beanbag44: yeah, all requests are expected to be sent on a tickly basis to stay up to date with the rest of the game

src/main/kotlin/com/lambda/module/modules/player/InventoryCleaner.kt
```diff
@@ -0,0 +1,130 @@
...
+import net.minecraft.screen.slot.SlotActionType
+
+object InventoryCleaner : Module(
+	name = "InventoryCleaner",
+	description = """
+		Automatically drops unwanted items when the player inventory is full.
+		""".trimIndent(),
+	tag = ModuleTag.PLAYER
+) {
+	private val itemsCanTrash by setting(
```
beanbag44: settings should be written on one line each

src/main/kotlin/com/lambda/module/modules/player/InventoryCleaner.kt
```diff
@@ -0,0 +1,130 @@
...
+						pickup(ScreenHandler.EMPTY_SPACE_SLOT_INDEX, 0)
+					}.submit(true)
+					event.cancel()
+				}
+			}
+		}
+	}
+
+	fun SafeContext.checkShouldTrashSomething() {
+		if (player.health > 0.0f && !player.isSpectator && dropToPickup && !player.isCreative) {
```
beanbag44: could probably replace the health check for player.isAlive. I'd also just check if the players in survival

beanbag44 (changes requested):
when clicking an empty slot, it still throws a stack to make space

src/main/kotlin/com/lambda/module/modules/player/InventoryCleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+	name = "InventoryCleaner",
+	description = """
+		Automatically drops unwanted items when the player inventory is full.
+		""".trimIndent(),
+	tag = ModuleTag.PLAYER
+) {
+	private val itemsCanTrash by setting("Items Can Trash", setOf<Item>(Items.NETHERRACK, Items.COBBLESTONE), ITEM.toSet(), description = "A list of items that the module can trash when trying to make space in the inventory.")
+
+	private val dropOnContainerClick by setting("Drop On Container Click", true, description = "If enabled, shift + left clicking an item in a container will drop a trash item out of the inventory if the player inventory is full.")
+	private val dropToPickup by setting(
```
beanbag44: setting not inlined

src/main/kotlin/com/lambda/module/modules/player/InventoryCleaner.kt
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	fun SafeContext.dropOneTrashStack(): Boolean {
+		StackSelection.selectStack {
+			isOneOfItems(itemsCanTrash)
+		}.filterSlots(InventoryContainer.slots).firstOrNull()?.let {
+			inventoryRequest {
+				throwStack(it.id)
+			}.submit(true)
```
beanbag44, line 114: prob best to just return if the inventory request here is done

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/player/InventoryCleaner.kt
```diff
@@ -0,0 +1,119 @@
...
+import net.minecraft.screen.GenericContainerScreenHandler
+import net.minecraft.screen.ScreenHandler
+import net.minecraft.screen.slot.SlotActionType
+import net.minecraft.world.GameMode
+
+@Suppress("unused")
+object InventoryCleaner : Module(
+	name = "InventoryCleaner",
+	description = """
+		Automatically drops unwanted items when the player inventory is full.
```
beanbag44: doesnt need the """

beanbag44 (changes requested):

src/main/kotlin/com/lambda/event/events/InventoryEvent.kt
```diff
@@ -82,6 +84,33 @@ sealed class InventoryEvent {
         val stack: ItemStack,
     ) : Event
 
+	/**
+	 * Represents an action performed on an inventory slot, such as clicking or interacting with it.
+	 */
+	abstract class SlotAction : Event, ICancellable {
```
emyfops: Should be open, also what is up with all this ktdoc

IceTank: Just copied what the constructor did in the rest of the file

Issue #206, opened 2025-12-21
Feat: Nametag and Tracers


JIexaproger:
### Feature

I would like to implement a feature Nametag and Tracers

### Pull Request

- [x] I will be submitting a pull request

emyfops:
Tracers and nametags are already being added by constructor in #198

PR #230, opened 2026-01-27
Improvement: Renders

beanbag44:
Largely reworks the rendering system and advances the api capabilities.

- SDF Font rendering
- item image rendering like in inventories
- world / screen renders like boxes, lines, quads, text. All have both variants aside from items currently
- keeps the original 3 renderer system (immediate, ticked, and chunked)


emyfops:
closes #206

PR #233, opened 2026-01-28
Feat: Add auto updater module

THCFree:
### Description
Adds Auto Updater module which installs Lambda loader which automatically loads latest version of lambda client depending on the selected release channel also uninstalls the loader and replaces it with the latest version of lambda for the current version when disabled. The module prompts the user trough a modal before installing / uninstalling since the process requires a client restart


master7720:
i wonder if u could impl so it auto updates to newest commits rather than version release ob there should be a setting between both

THCFree:
> i wonder if u could impl so it auto updates to newest commits rather than version release ob there should be a setting between both

If you switch client branch from stable to snapshot it will update to latest commit on every launch given that the commit is on your Minecraft version and the build completed correctly

PR #234, opened 2026-02-01
feat(AutoSpiral): Add AutoSpiral module

IceTank:
Add AutoSpiral module that works in the overworld like worlds and nether


src/main/kotlin/com/lambda/module/modules/movement/autospiral/SpiralIterator2d.kt
```diff
@@ -0,0 +1,89 @@
...
+        this.i = 0
+        this.j = 0
+        this.segment_passed = 0
+        this.k = 0
+    }
+
+    override fun next(): BlockPos? {
+        if (this.k >= this.NUMBER_OF_POINTS) return null
+        val output = BlockPos(this.i, 0, this.j)
+        this.i += this.di
```
KMatias123: These variable names could be more descriptive

IceTank: If I knew what those values did, I would give them better names. Ima ask AI to rename them.

beanbag44: bruh, at least understand what your code does before publishing it. AI is helpful but only when the code it writes is understood and verified to be up to standard

IceTank: > bruh, at least understand what your code does before publishing it. AI is helpful but only when the code it writes is understood and verified to be up to standard

The code was originally from Stackoverflow when I added it to Echelon. I used it in Echelon a lot without issues, so I just copy and pasted the code over. I think I wrote that code before AI was even a thing. That was one of the first modules I made.

emyfops:
You shouldn't add folders inside the modules folder unless it's a new category

src/main/kotlin/com/lambda/config/Configurable.kt
```diff
@@ -102,6 +108,12 @@ abstract class Configurable(
         }
     }
 
+    fun button(
```
beanbag44: I don't think this is the best way to go about adding non-setting elements in the module gui. They seem to be hard coded to appear after non-grouped settings and before grouped settings and dont store any data in the config.

It would be nice to add arbitrary imgui elements in modules for more unique functionality, but this would likely rely on some sort of switch up in how configurables store data in the config jsons

src/main/kotlin/com/lambda/module/modules/movement/AutoSpiral.kt
```diff
@@ -0,0 +1,115 @@
...
+	var currentWaypoint: BlockPos? = null
+	var center: BlockPos = BlockPos.ORIGIN
+
+	var spiralSpacing by setting("Spiral Spacing", 128, 16..1024, description = "The distance between each loop of the spiral")
+	var waypointTriggerDistance by setting("Waypoint Trigger Distance", 4, 2..64, description = "The distance to the waypoint at which a new waypoint is generated. Put in 50-60 range when in the Nether.")
+	var setCenterOnEnable by setting("Set Center On Enable", true, description = "Whether to set the center of the spiral to your current position when enabling the module.")
+	var setBaritoneGoal by setting("Set Baritone Goal", true, description = "Whether to set Baritone's goal to the current waypoint. Mostly so you can see where the next waypoint is.")
+
+	init {
+		button("Reset Center") {
```
beanbag44: for now i'd remove these and replace them with block pos settings

src/main/kotlin/com/lambda/interaction/managers/rotating/RotationRequest.kt
```diff
@@ -145,6 +145,9 @@ interface IRotationRequest : Automated {
 		@RotationRequestDsl
 		fun yaw(yaw: Float) { yawBuilder = { yaw.toDouble() } }
 
+		@RotationRequestDsl
+		fun yaw(rotation: Rotation) { yawBuilder = { rotation.yaw } }
```
beanbag44: I dont think theres any need for this and the other pitch function. Best just to call yaw() with rotation.yaw

src/main/kotlin/com/lambda/interaction/BaritoneManager.kt
```diff
@@ -365,11 +366,25 @@ object BaritoneManager : Configurable(LambdaConfig), Automated by AutomationConf
         primary?.customGoalProcess?.setGoalAndPath(goal)
     }
 
+    /**
+     * Sets the current Baritone goal without starting pathing
+     */
+    fun setGoal(goal: Goal) {
+        if (!isBaritoneLoaded || primary?.elytraProcess?.isLoaded == false) return
```
beanbag44: would it be better to do primary?.elytraProcess?.isLoaed != true? That way it would account for if its null too

src/main/kotlin/com/lambda/interaction/BaritoneManager.kt
```diff
@@ -365,11 +366,25 @@ object BaritoneManager : Configurable(LambdaConfig), Automated by AutomationConf
...
+	    primary?.customGoalProcess?.goal = goal
+    }
+
+    fun setGoalAndElytraPath(goal: Goal) {
+        if (!isBaritoneLoaded || primary?.elytraProcess?.isLoaded == false) return
```
beanbag44: same here

src/main/kotlin/com/lambda/util/BlockPosIterators.kt
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+package com.lambda.util
+
+import net.minecraft.util.math.BlockPos
+import kotlin.math.floor
+import kotlin.math.pow
+
+/**
+ * A collection of Block position iterator implementations for various purposes.
```
beanbag44, line 25: small nitpick but theres only one implementation, not a collection 🤓

src/main/kotlin/com/lambda/module/modules/movement/AutoSpiral.kt
```diff
@@ -0,0 +1,115 @@
...
+import com.lambda.util.extension.isNether
+import net.minecraft.util.math.BlockPos
+import kotlin.math.sqrt
+
+object AutoSpiral : Module(
+	name = "AutoSpiral",
+	description = "Automatically flies in a spiral pattern. Uses Baritone elytra pathing in the Nether.",
+	tag = ModuleTag.Companion.MOVEMENT,
+) {
+	var iterator: BlockPosIterators.SpiralIterator2d? = null
```
beanbag44, line 41: might as well make all these settings private

beanbag44 (changes requested):
cool module, some nitpicks

src/main/kotlin/com/lambda/util/BlockPosIterators.kt
IceTank: There might be in the future like a 3d version or some other pattern

IceTank:
Not to sure what to do about the buttons

beanbag44:
for now just replace them with blockpos settings

IceTank:
I removed the GuiButton implementaiton

src/main/kotlin/com/lambda/module/modules/movement/AutoSpiral.kt
```diff
@@ -0,0 +1,107 @@
...
+
+		listen<TickEvent.Pre> {
+			if (currentWaypoint == null || waypointReached()) {
+				nextWaypoint()
+			}
+
+			currentWaypoint?.let { waypoint ->
+				if (!world.isNether) {
+					rotationRequest {
+						lookAt(waypoint.toCenterPos()).yaw
```
beanbag44: rotation needs to be set in the request for it to be valid. This currently crashes my game

src/main/kotlin/com/lambda/module/modules/movement/AutoSpiral.kt
```diff
@@ -0,0 +1,107 @@
...
+					}.submit(true)
+				}
+			}
+		}
+	}
+
+	private fun SafeContext.waypointReached(): Boolean {
+		return currentWaypoint?.let {
```
beanbag44: can remove the return here and change the function into a basic = rather than a block

src/main/kotlin/com/lambda/module/modules/movement/AutoSpiral.kt
```diff
@@ -0,0 +1,107 @@
...
+	var setBaritoneGoal by setting("Set Baritone Goal", true, description = "Whether to set Baritone's goal to the current waypoint. Mostly so you can see where the next waypoint is.")
+
+	var center by setting("Center", BlockPos.ORIGIN, description = "Center position for the spiral")
+
+	init {
+		onEnable {
+			if (iterator == null) {
+				iterator = BlockPosIterators.SpiralIterator2d(10000)
+				if (setCenterOnEnable) {
+					center = BlockPos.ORIGIN
```
beanbag44: cant test, but why not set this to the players position?

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/movement/AutoSpiral.kt
IceTank: Feature

src/main/kotlin/com/lambda/module/modules/movement/AutoSpiral.kt
IceTank: yeah

PR #236, opened 2026-02-01
feat(AutoSign): Add AutoSign Module

IceTank:
Add AutoSign Module to autofill sign texts and auto close ópened signs


src/main/resources/lambda.accesswidener
```diff
@@ -123,3 +123,6 @@ transitive-accessible method net/minecraft/item/BlockItem getPlacementState (Lne
 transitive-accessible method net/minecraft/block/AbstractBlock getPickStack (Lnet/minecraft/world/WorldView;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/BlockState;Z)Lnet/minecraft/item/ItemStack;
 transitive-accessible field net/minecraft/client/gui/screen/ingame/HandledScreen focusedSlot Lnet/minecraft/screen/slot/Slot;
 transitive-accessible field net/minecraft/registry/SimpleRegistry frozen Z
+
+# AutoSign / GuiEvent.SignEditorOpen event
+accessible field net/minecraft/client/gui/screen/ingame/AbstractSignEditScreen messages [Ljava/lang/String;
```
emyfops: No need for the comment and ALWAYS MAKE THEM TRANSITIVE

src/main/resources/lambda.accesswidener
```diff
@@ -123,3 +123,5 @@ transitive-accessible method net/minecraft/item/BlockItem getPlacementState (Lne
 transitive-accessible method net/minecraft/block/AbstractBlock getPickStack (Lnet/minecraft/world/WorldView;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/BlockState;Z)Lnet/minecraft/item/ItemStack;
 transitive-accessible field net/minecraft/client/gui/screen/ingame/HandledScreen focusedSlot Lnet/minecraft/screen/slot/Slot;
 transitive-accessible field net/minecraft/registry/SimpleRegistry frozen Z
+
+transitive-accessible field net/minecraft/client/gui/screen/ingame/AbstractSignEditScreen messages [Ljava/lang/String;
```
beanbag44, line 157: prob just remove the white space above to move it into the # Other group

src/main/kotlin/com/lambda/module/modules/player/AutoSign.kt
```diff
@@ -0,0 +1,97 @@
...
+
+					if (formattedLine.isNotEmpty()) lines[i] = String.format(formattedLine, Date())
+				}
+			}
+
+			var editor: AbstractSignEditScreen = if (event.sign is HangingSignBlockEntity) HangingSignEditScreen(event.sign, true, mc.shouldFilterText())
+			else SignEditScreen(event.sign, true, mc.shouldFilterText())
+			for (i in 0 until 4) editor.messages[i] = lines[i]
+			if (autoClose) {
+				if (writeOnFront) mc.networkHandler?.sendPacket(UpdateSignC2SPacket(event.sign.pos, true, editor.messages[0], editor.messages[1], editor.messages[2], editor.messages[3]))
```
beanbag44: SafeContext contains connection which should be used instead of mc.networkHandler. Also 2b2t doesnt allow sign editing so this doesnt seem to work. Would it be possible to close the sign automatically without using edit sign?

beanbag44 (changes requested):
looks good, couple changes though

beanbag44:
auto close still doesnt work on 2b because sign editing is disabled. Also dw about using @Suppress unused. No other module does that so might as well keep it consistent

src/main/kotlin/com/lambda/module/modules/player/AutoSign.kt
IceTank: Now works on 2b

PR #238, opened 2026-02-02
feat(AutoWalk): Add basic AutoWalk module

IceTank:
Add basic AutoWalk module

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
@@ -35,8 +35,6 @@ class AutoWalk : Module(
 	description = "Automatically makes your character walk forward",
 	tag = ModuleTag.MOVEMENT,
 ) {
-	val pause by setting("Enable Pause", false)
-
 	val limitSpeed by setting("Limit Speed", false)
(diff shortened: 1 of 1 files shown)


IceTank:
Not tested

IceTank:
Works

src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.util.math.Vec2f
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
+
+class AutoWalk : Module(
+	name = "AutoWalk",
+	description = "Automatically makes your character walk forward",
+	tag = ModuleTag.MOVEMENT,
+) {
+	val pause by setting("Enable Pause", false)
```
beanbag44: this setting doesnt seem to be used?

beanbag44 (changes requested):
good but unused setting

src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
IceTank: Forgot to remove that

beanbag44 (approved):
nice

PR #240, opened 2026-02-02
feat(AutoMount): Add AutoMount module

IceTank:
Add AutoMount module to automatically mount or remount entities

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2026 Lambda
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.lambda.module.modules.movement
+
+import com.lambda.context.SafeContext
+import com.lambda.event.events.TickEvent
+import com.lambda.event.listener.SafeListener.Companion.listen
+import com.lambda.module.Module
+import com.lambda.module.tag.ModuleTag
+import com.lambda.util.Timer
+import com.lambda.util.world.fastEntitySearch
+import net.minecraft.entity.Entity
+import net.minecraft.entity.EntityType
+import net.minecraft.registry.Registries
+import kotlin.time.Duration.Companion.milliseconds
+
+class AutoMount : Module(
+	name = "AutoMount",
+	description = "Automatically mounts entities",
+	tag = ModuleTag.MOVEMENT
+) {
+	var autoRemount by setting("Auto Remount", false, description = "Automatically remounts if you get off")
+	var autoMountEntities by setting("Auto Mount Entities", true, description = "Automatically mounts nearby entities in range")
+	var autoMountEntityList by setting("Auto Mount Entity List", mutableListOf<EntityType<*>>(), mutableListOf(Registries.ENTITY_TYPE.toList())) { autoMountEntities }
+
+	var interval by setting("Interval", 50, 1..200, 1, unit = "ms", description = "Interact interval")
+	var range by setting("Range", 5.0, 1.0..10.0, 0.1, description = "Mount range")
+
+	val intervalTimer = Timer()
+	var lastEntity: Entity? = null
+
+	init {
+		listen<TickEvent.Pre> {
+			if (!intervalTimer.timePassed(interval.milliseconds)) {
+				return@listen
+			}
+			if (autoMountEntities && !player.isRiding) {
+				val entity = fastEntitySearch<Entity>(range) {
+					autoMountEntityList.contains(it.type) && canRide(it) && it.distanceTo(player) <= range
+				}
+				entity.firstOrNull()?.let {
+					intervalTimer.reset()
+					player.startRiding(it)
+				}
+			}
+			if (autoRemount) {
+				if (player.isRiding) {
+					lastEntity = player.vehicle
+				} else {
+					lastEntity?.let {
+						if (it.isRemoved || it.distanceTo(player) > range) {
+							lastEntity = null
+						} else {
+							if (canRide(it)) {
+								intervalTimer.reset()
+								player.startRiding(it)
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+
+	private fun SafeContext.canRide(entity: Entity): Boolean {
+		return entity.canAddPassenger(player)
+	}
+
+	private fun Entity.canAddPassenger(other: Entity): Boolean {
+		return this.canAddPassenger(other)
+	}
+}
--- a/src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
@@ -17,16 +17,27 @@
 
 package com.lambda.module.modules.movement
 
+import com.lambda.config.AutomationConfig
 import com.lambda.context.SafeContext
 import com.lambda.event.events.TickEvent
 import com.lambda.event.listener.SafeListener.Companion.listen
+import com.lambda.interaction.managers.rotating.visibilty.VisibilityChecker
+import com.lambda.interaction.managers.rotating.visibilty.VisibilityChecker.findRotation
 import com.lambda.module.Module
 import com.lambda.module.tag.ModuleTag
+import com.lambda.threading.runSafeAutomated
+import com.lambda.util.Communication.debug
+import com.lambda.util.Communication.info
+import com.lambda.util.EntityUtils
 import com.lambda.util.Timer
+import com.lambda.util.math.dist
 import com.lambda.util.world.fastEntitySearch
 import net.minecraft.entity.Entity
 import net.minecraft.entity.EntityType
+import net.minecraft.network.packet.c2s.play.PlayerInteractEntityC2SPacket
 import net.minecraft.registry.Registries
+import net.minecraft.util.Hand
+import net.minecraft.util.math.Vec3d
 import kotlin.time.Duration.Companion.milliseconds
 
 class AutoMount : Module(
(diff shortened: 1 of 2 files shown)


IceTank:
Not tested yet

IceTank:
Works now

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+						autoMountEntityList.contains(it.type) && canRide(it) && it.findRotation(range, player.eyePos) != null
+					}.sortedBy { it.squaredDistanceTo(player.pos) }
+					entity.firstOrNull()?.let {
+						intervalTimer.reset()
+						interactEntity(it)
+						debug("Mounting ${it.name}")
+					}
+				}
+			}
+			if (autoRemount) {
```
beanbag44: in situations like this i would check if (!autoRemount) return@listen

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+				if (player.vehicle != null) {
+					lastEntity = player.vehicle
```
beanbag44: here for example, you could add return@listen in the if statement and then you dont have to wrap the rest in the else

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+				} else {
+					lastEntity?.let {
+						if (it.isRemoved || it.distanceTo(player) > range) {
+							lastEntity = null
```
beanbag44: here as well; return@let

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+	var autoMountEntities by setting("Auto Mount Entities", true, description = "Automatically mounts nearby entities in range")
+	var autoMountEntityList by setting("Auto Mount Entity List",
+		mutableListOf(),
+		Registries.ENTITY_TYPE.toList()
+	) { autoMountEntities }
+
+	var interval by setting("Interval", 50, 1..200, 1, unit = "ms", description = "Interact interval")
+	var range by setting("Range", 4.0, 1.0..20.0, 0.1, description = "Mount range")
+
+	override var automationConfig = AutomationConfig(
```
beanbag44: automation configs are set using setDefaultAutomationConfig at the start of the init block

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+import kotlin.time.Duration.Companion.milliseconds
+
+class AutoMount : Module(
+	name = "AutoMount",
+	description = "Automatically mounts entities",
+	tag = ModuleTag.MOVEMENT
+) {
+	var autoRemount by setting("Auto Remount", false, description = "Automatically remounts if you get off")
+	var autoMountEntities by setting("Auto Mount Entities", true, description = "Automatically mounts nearby entities in range")
+	var autoMountEntityList by setting("Auto Mount Entity List",
```
beanbag44: settings should be written on one line with the exception of some builder style functions like onSelect, onDeselect, onValueChange, etc

src/main/resources/lambda.accesswidener
```diff
@@ -123,3 +123,6 @@ transitive-accessible method net/minecraft/item/BlockItem getPlacementState (Lne
 transitive-accessible method net/minecraft/block/AbstractBlock getPickStack (Lnet/minecraft/world/WorldView;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/BlockState;Z)Lnet/minecraft/item/ItemStack;
 transitive-accessible field net/minecraft/client/gui/screen/ingame/HandledScreen focusedSlot Lnet/minecraft/screen/slot/Slot;
 transitive-accessible field net/minecraft/registry/SimpleRegistry frozen Z
+
+# AutoMount
+accessible method net/minecraft/entity/Entity canAddPassenger (Lnet/minecraft/entity/Entity;)Z
```
beanbag44: this can just be placed under the # Entity group

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		listen<TickEvent.Pre> {
+			if (!intervalTimer.timePassed(interval.milliseconds)) {
+				return@listen
+			}
+			if (autoMountEntities && player.vehicle == null) {
+				runSafeAutomated {
+					val entity = fastEntitySearch<Entity>(10.0) {
+						autoMountEntityList.contains(it.type) && canRide(it) && it.findRotation(range, player.eyePos) != null
```
beanbag44, line 82: I think its best to have a rotation setting as currently this checks if its possible for a valid rotation but doesnt perform it.

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+					}.sortedBy { it.squaredDistanceTo(player.pos) }
+					entity.firstOrNull()?.let {
+						intervalTimer.reset()
+						interactEntity(it)
+						debug("Mounting ${it.name}")
```
beanbag44: just a nitpick but a log setting or something could be good

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+
+	private fun SafeContext.interactEntity(entity: Entity) {
+		mc.networkHandler?.sendPacket(PlayerInteractEntityC2SPacket.interactAt(entity, false, Hand.MAIN_HAND, Vec3d(0.5, 0.5, 0.5)))
```
beanbag44: ideally we should fill the parameters with accurate data like player.isSneaking instead of false

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,114 @@
...
+		mc.networkHandler?.sendPacket(PlayerInteractEntityC2SPacket.interact(entity, false, Hand.MAIN_HAND))
+	}
+
+	private fun SafeContext.canRide(entity: Entity): Boolean {
```
beanbag44: can be inlined into
```kotlin
private fun SafeContext.canRide(entity: Entity) = entity.canAddPassenger(player)
```

beanbag44 (changes requested):

beanbag44:
oh yeah, also use connection inside of SafeContext rather than mc.networkHandler

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
IceTank: That is only to check if the entity is within range as all the other checks I tried didn't work

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
IceTank: This would cause the module to fail to mount entities while the player is sneaking

beanbag44: could this potentially cause anticheat conflicts tho? Maybe a setting for override sneak or something

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
beanbag44: yeah but this module wont work on grim tho? Might as well have an option for rotations as you already have the logic for getting the rotation

src/main/resources/lambda.accesswidener
beanbag44: this still could be moved

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
IceTank: It does work on grim. Grim does not seam to check rotations when mounting entities

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
IceTank: Idk works on 2b

src/main/resources/lambda.accesswidener
```diff
@@ -62,6 +62,7 @@ transitive-accessible field net/minecraft/entity/Entity world Lnet/minecraft/wor
 transitive-accessible class net/minecraft/screen/slot/ArmorSlot
 transitive-accessible field net/minecraft/screen/slot/ArmorSlot equipmentSlot Lnet/minecraft/entity/EquipmentSlot;
 transitive-accessible field net/minecraft/entity/Entity FLAGS Lnet/minecraft/entity/data/TrackedData;
+accessible method net/minecraft/entity/Entity canAddPassenger (Lnet/minecraft/entity/Entity;)Z
```
beanbag44: add transitive- at the beginning

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+	var interval by setting("Interval", 50, 1..200, 1, unit = "ms", description = "Interact interval")
+	var range by setting("Range", 4.0, 1.0..20.0, 0.1, description = "Mount range")
+	var rotate by setting("Rotate", false, description = "Rotate to the entity when mounting")
+	var debug by setting("Debug", false, description = "Print debug messages")
+
+	val intervalTimer = Timer()
+	var lastEntity: Entity? = null
+
+	init {
+		setDefaultAutomationConfig("AutoMount")
```
beanbag44: best to hide unused configs in the automation config

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+import kotlin.time.Duration.Companion.milliseconds
+
+class AutoMount : Module(
+	name = "AutoMount",
+	description = "Automatically mounts entities",
+	tag = ModuleTag.MOVEMENT
+) {
+	var autoRemount by setting("Auto Remount", false, description = "Automatically remounts if you get off")
+	var autoMountEntities by setting("Auto Mount Entities", true, description = "Automatically mounts nearby entities in range")
+	var autoMountEntityList by setting("Auto Mount Entity List", mutableListOf(), Registries.ENTITY_TYPE.toList()) { autoMountEntities }
```
beanbag44, line 52: we have an entity selection settings group now which you could use and modify the default selected

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+
+	private fun SafeContext.interactEntity(entity: Entity) {
+		if (rotate) {
+			runSafeAutomated {
+				var hit: VisibilityChecker.CheckedHit? = null
+				rotationRequest {
+					hit = lookAtEntity(entity)
+				}.submit()
```
beanbag44: no check for if the rotation is done

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+			}
+		}
+	}
+
+	private fun SafeContext.interactEntity(entity: Entity) {
+		if (rotate) {
+			runSafeAutomated {
+				var hit: VisibilityChecker.CheckedHit? = null
+				rotationRequest {
+					hit = lookAtEntity(entity)
```
beanbag44: this doesnt actually set the rotation in the request. Best to set the hit value before the request, then set the rotation in the request

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+				}
+				if (canRide(it)) {
+					intervalTimer.reset()
+					interactEntity(it)
+				}
+			}
+		}
+	}
+
+	private fun SafeContext.interactEntity(entity: Entity) {
```
beanbag44: could make the whole function AutomatedSafeContext rather than SafeContext as one of the call sites already has an automated safe context and it would cut down on an extra indent

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+					return@runSafeAutomated
+				} ?: run {
+					if (debug) debug("Not rotation found to mount entity")
+					return@runSafeAutomated
+				}
+				if (debug) debug("Invalid entity rotation")
+			}
+		} else {
+			connection.sendPacket(PlayerInteractEntityC2SPacket.interactAt(entity, false, Hand.MAIN_HAND, Vec3d(0.5, 0.5, 0.5)))
+			connection.sendPacket(PlayerInteractEntityC2SPacket.interact(entity, false, Hand.MAIN_HAND))
```
beanbag44, line 134: isnt this only sent if the first one isnt successful in vanilla mc?

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+	private fun SafeContext.interactEntity(entity: Entity) {
+		if (rotate) {
+			runSafeAutomated {
+				var hit: VisibilityChecker.CheckedHit? = null
+				rotationRequest {
+					hit = lookAtEntity(entity)
+				}.submit()
+				hit?.let {
+					val hitResult = it.hit as? EntityHitResult ?: return@let
+					interaction.interactEntityAtLocation(player, entity, hitResult, Hand.MAIN_HAND)
```
beanbag44: this doesnt have the player sneaking override to set it to false

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
```diff
@@ -0,0 +1,126 @@
...
+			if (player.vehicle?.isRemoved == false) {
+				lastEntity = player.vehicle
+			}
+			lastEntity?.let {
+				if (it.isRemoved || it.distanceTo(player) > range) {
+					lastEntity = null
+					return@let
+				}
+				if (canRide(it)) {
+					intervalTimer.reset()
```
beanbag44: could prob put this at the start of interactEntity as its called before every invocation

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
IceTank: idk

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
IceTank: What is that called?

IceTank:
Re BASED and updated and fixed some code from pr comments

src/main/kotlin/com/lambda/module/modules/movement/AutoMount.kt
beanbag44: EntitySelectionSettings

IceTank:
@beanbag44 any updates here?

beanbag44:
theres still the couldAcceptPassenger thing

PR #241, opened 2026-02-04
List selection popup Model

IceTank:
Adds a popup model to select items from. It should be a lot easier to work with than having to scroll through the dropdown list.

image of the List selection popup model featuring two selection menues to select entries from

You can make selections on either side of item lists and then change the side of the selected items with the arrow buttons. The "X" next to the search bar clears the search entry. "Select all" selects all currently visible items. The search field does not auto-clear when the model closes. This is intended behavior.

Code is kinda popo atm. I'll have to figure out some better ways to create the model window.


IceTank:
I guess this is already a thing now

PR #247, opened 2026-02-27
Freecam improvements

IceTank:
Adds a follow mode to freecam that keeps the freecam camera close to the player but only moves it if the player is moving away. This also adds an option in follow mode to track the player's eye position with the camera.

https://github.com/user-attachments/assets/03c5c7ef-ed2a-4d12-b937-3b269c9578ec


src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +50,206 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+
+	private fun Vec2d.distanceTo(other: Vec2d): Double {
+		val dx = other.x - this.x
+		val dz = other.y - this.y
+		return kotlin.math.sqrt(dx * dx + dz * dz)
+	}
+
+	private fun Vec2d.normalize(): Vec2d {
+		val length = distanceTo(Vec2d(0.0, 0.0))
+		return if (length > 0) Vec2d(x / length, y / length) else Vec2d(0.0, 0.0)
```
beanbag44: could use Vec2d.ZERO

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +50,206 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+
+	private fun SafeContext.findFollowTarget(): PlayerEntity? {
+		if (player.gameMode != GameMode.SPECTATOR) {
+			return player
+		}
+		val players = world.players.filter { it != player && it !is ClientPlayerEntity }
+		return players.minByOrNull { it.eyePos.squaredDistanceTo(position) }
+	}
+
+	private fun Vec2d.distanceTo(other: Vec2d): Double {
```
beanbag44: we have the dist functions in Vectors.kt, could prob just add the extension functions for dist and distSq in there for Vec2d

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +50,206 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+
+	enum class FreecamMode(override val displayName: String, override val description: String) : NamedEnum, Describable {
+		Free("Free", "Move the camera freely with keyboard input"), FollowPlayer("Follow Player", "Camera follows a player as if attached by an invisible string");
+	}
+
+	private fun SafeContext.findFollowTarget(): PlayerEntity? {
+		if (player.gameMode != GameMode.SPECTATOR) {
+			return player
+		}
+		val players = world.players.filter { it != player && it !is ClientPlayerEntity }
```
beanbag44: could remove the it != player check as the ClientPlayerEntity check already acounts for that

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +50,206 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+			}
+		}
+
+		listen<RenderEvent.UpdateTarget>(priority = 1) { event -> // Higher priority then RotationManager to run before RotationManager modifies mc.crosshairTarget
+			mc.crosshairTarget = rotation.rayCast(reach, lerpPos).orMiss // Can't be null (otherwise mc will spam "Null returned as 'hitResult', this shouldn't happen!")
+			mc.crosshairTarget?.let { if (it.type != HitResult.Type.MISS) event.cancel() }
+		}
+	}
+
+	enum class FreecamRotationMode(override val displayName: String, override val description: String) : NamedEnum, Describable {
```
beanbag44: usually for local private enums i dont bother with the name prefix and just use RotationMode or Mode for example. Also these could be private

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +50,206 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+	private val relative by setting("Relative", false, "Moves freecam relative to player position") { mode == FreecamMode.Free }.onValueChange { _, it -> if (it) lastPlayerPosition = player.pos }
+
+	// Follow Player settings
+	private val followMaxDistance by setting("String Length", 10.0, 2.0..50.0, 0.5, "Maximum distance before the string pulls the camera", unit = "m") { mode == FreecamMode.FollowPlayer }
+	private val followTrackPlayer by setting("Track Player", false, "Keeps looking at the followed player") { mode == FreecamMode.FollowPlayer }
+
+	private val keepYLevel by setting("Keep Y Level", false, "Don't change the camera y-level on player movement. Applies to relative and follow modes")
+
+
+	override val rotationConfig = RotationConfig.Instant(RotationMode.Lock)
```
beanbag44: this should be applied using setDefaultAutomationConfig { hideAllGroupsExcept(rotationConfig) } or something like that in the init block before the listeners

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +50,206 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+		onDisable {
+			mc.options.perspective = lastPerspective
+		}
+
+		listen<TickEvent.Pre> {
+			when (rotateMode) {
+				FreecamRotationMode.None -> return@listen
+				FreecamRotationMode.KeepRotation -> rotationRequest { rotation(rotation) }.submit()
+				FreecamRotationMode.LookAtTarget -> mc.crosshairTarget?.let {
+					runSafeAutomated {
```
beanbag44: i dont think runSafeAutomated is required here

beanbag44 (changes requested):
really cool, didnt know it would act like a string attached to the player. Some changes tho

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +51,199 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+	/**
+	 * @see net.minecraft.entity.Entity.changeLookDirection
+	 */
+	private const val SENSITIVITY_FACTOR = 0.15
+
+	init {
+		setDefaultAutomationConfig {
+			applyEdits {
+				rotationConfig::rotationMode.edit {
+					defaultValue(com.lambda.interaction.managers.rotating.RotationMode.Lock)
```
beanbag44: full qualified reference. Best to add an import

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +51,199 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+				}
+				hideAllGroupsExcept(rotationConfig)
+				rotationConfig.apply {
+
```
beanbag44: empty rotationConfig.apply {}

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +51,199 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+	private val sprint by setting("Sprint Multiplier", 3.0, 0.1..10.0, 0.1, description = "Set below 1.0 to fly slower on sprint.") { mode == Mode.Free }
+	private val reach by setting("Reach", 10.0, 1.0..100.0, 1.0, "Freecam reach distance")
+	private val rotateMode by setting("Rotate Mode", RotationMode.None, "Rotation mode").onValueChange { _, it -> if (it == RotationMode.LookAtTarget) mc.crosshairTarget = BlockHitResult.createMissed(Vec3d.ZERO, Direction.UP, BlockPos.ORIGIN) }
+	private val relative by setting("Relative", false, "Moves freecam relative to player position") { mode == Mode.Free }.onValueChange { _, it -> if (it) lastPlayerPosition = player.pos }
+
+	// Follow Player settings
+	private val followMaxDistance by setting("String Length", 10.0, 2.0..50.0, 0.5, "Maximum distance before the string pulls the camera", unit = "m") { mode == Mode.FollowPlayer }
+	private val followTrackPlayer by setting("Track Player", false, "Keeps looking at the followed player") { mode == Mode.FollowPlayer }
+
+	private val keepYLevel by setting("Keep Y Level", false, "Don't change the camera y-level on player movement. Applies to relative and follow modes")
```
beanbag44: this doesnt seem to do anything in the follow player mode. Could add a visibility check for that

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -47,133 +51,199 @@ import com.lambda.util.player.MovementUtils.roundedForward
...
+		}
+
+		listen<RenderEvent.UpdateTarget>(priority = 1) { event -> // Higher priority then RotationManager to run before RotationManager modifies mc.crosshairTarget
+			mc.crosshairTarget = rotation.rayCast(reach, lerpPos).orMiss // Can't be null (otherwise mc will spam "Null returned as 'hitResult', this shouldn't happen!")
+			mc.crosshairTarget?.let { if (it.type != HitResult.Type.MISS) event.cancel() }
+		}
+	}
+
+	private enum class RotationMode(override val displayName: String, override val description: String) : NamedEnum, Describable {
+		None("None", "No rotation changes"), LookAtTarget("Look At Target", "Look at the block or entity under your crosshair"), KeepRotation("Keep Rotation", "Look in the same direction as the camera");
```
beanbag44: easier to read this if its on separate lines

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
IceTank: there is a name conflict for RotationMode but I can rename it

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
IceTank: yeah

beanbag44:
actually just checking again. The keep y level setting should probably only display if relative is enabled too?

beanbag44:
also the mode enum should be on separate lines too. After this i'll merge it

PR #250, opened 2026-03-07
Rename air and add description to air setting in printer

IceTank:
# Pull Request Guidelines

### This is a template, modify before submitting your PR

Ensure that your title is concise and has all the necessary information

**Examples:**
- `Feat: Add new block types`
- `Fix: Crash on startup`
- `Fix: GUI does not render`
- `Ref: Optimize rendering engine`
- `New pre-processor for pathfinding`
- `Fix inconsistency with the structure`

### Issue Link
If your PR addresses one or more issues, be sure to link them. Use appropriate keywords like `closes`, `fixes`, or `resolves` to automatically close the linked issues when the PR is merged.

#### **Examples:**
- `This PR fixes a memory leak in foo() and closes #4. It also resolves #5, which is a duplicate issue.`
- `Closes #1, resolves #2, and fixes #3`
- `Fixes #1`

### Description
Provide a concise yet detailed summary of the changes introduced in this PR. Include the purpose of the changes and any relevant context.

#### **Examples:**
- `This pull request improves compatibility with version 1.21.4 by addressing rendering issues and adding support for new block types.`
- `Refactors the rendering engine to enhance performance and reduce memory usage.`
- `Adds a new feature for biome-specific block spawning to align with gameplay mechanics introduced in version 1.20.4.`

### Checklist Before Submitting
To ensure the quality and maintainability of your PR, confirm the following before submission:
1. Code adheres to the project's style guide and conventions.
2. All tests pass, including newly added tests.
3. Documentation has been updated to reflect any new features or changes.
4. Your PR is limited to a single purpose, avoiding unrelated changes.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/player/Printer.kt
+++ b/src/main/kotlin/com/lambda/module/modules/player/Printer.kt
@@ -18,10 +18,8 @@
 package com.lambda.module.modules.player
 
 import com.lambda.config.AutomationConfig.Companion.setDefaultAutomationConfig
-import com.lambda.config.applyEdits
 import com.lambda.interaction.construction.blueprint.TickingBlueprint
 import com.lambda.interaction.construction.verify.TargetState
-import com.lambda.interaction.managers.interacting.InteractConfig
 import com.lambda.module.Module
 import com.lambda.module.tag.ModuleTag
 import com.lambda.task.RootTask.run
@@ -39,7 +37,7 @@ object Printer : Module(
 	tag = ModuleTag.PLAYER
 ) {
 	private val range by setting("Range", 5, 1..7, 1)
-	private val air by setting("Air", false)
+	private val considerAll by setting("Consider All", false, description = "Consider all blocks outside the schematic as schematic air blocks")
 
 	private var buildTask: Task<*>? = null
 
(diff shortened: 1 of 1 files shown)


beanbag44:
i think "Air" makes more sense as its the most accurate. All blocks outside of schematics are considered air, and if air is disabled, it ignores air in the schematic too. Maybe just a better description but keep the original name?

IceTank:
Added another check to only interact with blocks that are inside any loaded schematic

IceTank:
I added another check to only consider blocks inside active placements for printing. This also prevents the printer from breaking shit outside the schematic. Air setting now breaks any blocks that do not match the schematic and should be air.

src/main/kotlin/com/lambda/module/modules/player/Printer.kt
```diff
@@ -66,6 +64,14 @@ object Printer : Module(
 		onDisable { buildTask?.cancel(); buildTask = null }
 	}
 
+	private fun inSchematic(pos: BlockPos): Boolean {
+		val placementManager = DataManager.getSchematicPlacementManager()
+		placementManager?.getAllPlacementsTouchingChunk(pos)?.forEach {
+			if (it.placement.isEnabled && it.bb.containsPos(pos)) return true
```
beanbag44: could use .any {} instead of returning true in the for loop and then returning false at the end if it reaches that point

beanbag44 (changes requested):
really nice, been wanting to fix that for a while. Just one small improvement

PR #258, opened 2026-03-07
feat(ModuleNotifier): Add Module to get feedback in chat when modules toggle

IceTank:
Adds chat feedback when modules toggle


src/main/kotlin/com/lambda/util/Communication.kt
```diff
@@ -96,18 +97,42 @@ object Communication {
...
-                player.sendMessage(log, false)
+                player.sendMessage(log, inGameOverlay)
+            }
+        }
+    }
+
+    /**
+     * Logs messages to the in game chat if available while keeping the message styles
+     */
+    fun Any.log(
```
beanbag44: im not sure this function is required. The function above takes a Text object which encompases MutableText too

src/main/kotlin/com/lambda/module/Module.kt
```diff
@@ -171,6 +178,22 @@ abstract class Module(
 
     fun toggle() {
         isEnabled = !isEnabled
+
+        if (ModuleNotifier.isEnabled) {
```
beanbag44: this would be better implemented using an event for module toggles. This also doesnt account for enable() and disable(). Perhaps to trigger the event you could add onEnable and onDisable callbacks. Then listen to the event in ModuleNotifier

src/main/kotlin/com/lambda/util/text/TextDsl.kt
```diff
@@ -120,6 +125,17 @@ fun TextBuilder.literal(value: String) {
...
 }
 
+/**
+ * Adds a literal text that is not re-styled by the builder's current style.
+ *
+ * @param value The text.
+ * @see StyleBuilder for action
+ */
+@TextDsl
+fun TextBuilder.literal(value: MutableText) {
```
beanbag44: if removing the extra function in Communication, this wont be required either

src/main/kotlin/com/lambda/util/text/TextDsl.kt
```diff
@@ -80,6 +80,11 @@ class TextBuilder {
         this.text.append(style.applyTo(text))
     }
 
+    @TextDsl
+    fun append(text: MutableText) {
```
beanbag44: this would also not be required

beanbag44 (changes requested):

IceTank:
I refactored the ModuleNotifier so it listens to events produced by the module when toggling

src/main/kotlin/com/lambda/util/Communication.kt
IceTank: Removed

src/main/kotlin/com/lambda/util/Communication.kt
```diff
@@ -39,6 +39,7 @@ import com.lambda.util.text.literal
 import com.lambda.util.text.styled
 import com.lambda.util.text.text
 import net.minecraft.client.toast.SystemToast
+import net.minecraft.text.MutableText
```
beanbag44: unused import

src/main/kotlin/com/lambda/util/Communication.kt
```diff
@@ -96,18 +97,22 @@ object Communication {
         }
     }
 
+    /**
+     * Logs messages to the in game chat if available while stripping the message styles
```
beanbag44: seems to be outdated now

src/main/kotlin/com/lambda/module/Module.kt
```diff
@@ -162,14 +168,23 @@ abstract class Module(
...
 
     fun disable() {
+        if (ModuleEvent.Disabled(this@Module).post().isCanceled()) {
+            return
+        }
         isEnabled = false
     }
 
     fun toggle() {
+        if (ModuleEvent.Toggle(this@Module, !isEnabled).post().isCanceled()) {
```
beanbag44: Enabled and Disabled dont seem to get called when the module is toggled. The toggle event should be posted first, then the corresponding event for the new value

src/main/kotlin/com/lambda/module/Module.kt
```diff
@@ -162,14 +168,23 @@ abstract class Module(
...
     fun enable() {
+        if (ModuleEvent.Enabled(this@Module).post().isCanceled()) {
+            return
+        }
         isEnabled = true
     }
 
     fun disable() {
+        if (ModuleEvent.Disabled(this@Module).post().isCanceled()) {
+            return
```
beanbag44: just a styling choice but i typically omit the brackets for early returns and small one-liner if blocks. The return can just be placed at the end of the if statement without the brackets, or just on the next line

src/main/kotlin/com/lambda/module/modules/client/ModuleNotifier.kt
```diff
@@ -0,0 +1,73 @@
...
+				val newState = if (it.newValue) "on" else "off"
+				val color = if (it.newValue) Colors.GREEN else Colors.RED
+				val message = Text.literal(newState).withColor(color)
+				logToTargets(message)
+			}
+		}
+	}
+
+	private fun logToTargets(message: Text) {
+		if (notifyTarget.contains(NotifyTarget.Chat)) {
```
beanbag44: could use *NotifyTarget.Chat in notifyTarget*

src/main/kotlin/com/lambda/module/modules/client/ModuleNotifier.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.lambda.util.Communication.log
+import com.lambda.util.Describable
+import com.lambda.util.NamedEnum
+import net.minecraft.text.Text
+import net.minecraft.util.Colors
+
+object ModuleNotifier : Module(
+	name = "ModuleNotifier",
+	description = "Notifies you when a module is enabled or disabled",
+	tag = com.lambda.module.tag.ModuleTag.CLIENT
```
beanbag44: use an import instead of the full qualified reference

src/main/kotlin/com/lambda/module/modules/client/ModuleNotifier.kt
```diff
@@ -0,0 +1,73 @@
...
+
+	enum class NotifyTarget(override val displayName: String, override val description: String) : Describable, NamedEnum {
+		Chat("Chat", "Sends a message to chat when a module is toggled"),
+		ActionBar("Action Bar", "Sends a message to the action bar when a module is toggled"),;
+	}
+
+	init {
+		listen<ModuleEvent.Enabled> {
+			runSafe {
+				logToTargets(Text.literal("on").withColor(Colors.GREEN))
```
beanbag44: we have a text dsl builder for stuff like this

beanbag44 (changes requested):
I dont know, just assuming but if portions of this pr and others are written with ai, please try to clean up the code (unused imports, old comments, etc), and test it to make sure it has the desired outcome

src/main/kotlin/com/lambda/module/Module.kt
IceTank: I think it's more readable this way. This way you can see the if statement at a glance and can expect the code flow to return or do some action within the code block. Otherwise, when glancing over, you might miss the return statement if it is at the end of the line.

src/main/kotlin/com/lambda/module/modules/client/ModuleNotifier.kt
```diff
@@ -0,0 +1,73 @@
...
+		listen<ModuleEvent.Disabled> { event ->
+			logToTargets(event.module, buildText {
+				color(Color(Colors.RED)) {
+					literal("off")
+				}
+			})
+		}
+	}
+
+	private fun logToTargets(module: Module, message: Text) {
```
beanbag44: could prob be changed to have the message: Text param be a TextBuilder.() -> Unit instead. That way you could have the block outside the param parenthesis

src/main/kotlin/com/lambda/module/modules/client/ModuleNotifier.kt
```diff
@@ -0,0 +1,73 @@
...
+import net.minecraft.text.Text
+import net.minecraft.util.Colors
+import java.awt.Color
+
+object ModuleNotifier : Module(
+	name = "ModuleNotifier",
+	description = "Notifies you when a module is enabled or disabled",
+	tag = ModuleTag.CLIENT
+) {
+	var notifyTarget by setting("Notify Target", setOf<NotifyTarget>(NotifyTarget.ActionBar), NotifyTarget.entries.toSet(), description = "Where to send notifications when modules are toggled")
```
beanbag44, line 41: i'd add params to the enum rather than using a collection setting here. One boolean for chat, and one for actionBar. Then have a third enum selection for Both, which sets true for both. Then just check if the enum has the value set to true for whatever youre testing

src/main/kotlin/com/lambda/event/events/ModuleEvent.kt
```diff
@@ -0,0 +1,55 @@
...
+
+package com.lambda.event.events
+
+import com.lambda.event.Event
+import com.lambda.event.callback.Cancellable
+import com.lambda.event.callback.ICancellable
+import com.lambda.module.Module
+
+/**
+ * Represents events related to toggling, enabling, and disabling of [Module]s.
```
beanbag44, line 26: the kt docs here are pretty self explanatory; can honestly just remove them for this class / child classes

beanbag44 (changes requested):
Im still not big on the idea of module toggles / on / off being cancellable. Maybe just remove that for now because i cant think of a time when that would be necessary. Aside from that as long as this is done its good to merge

beanbag44:
oh also prob good to make this module enabled by default as its pretty standard across clients

IceTank:
@beanbag44 merge this

beanbag44:
once i check it doesnt break everything and steal my bank info

IceTank:
You will never find the keylogger I added

src/main/kotlin/com/lambda/module/modules/client/ModuleNotifier.kt
IceTank: This is better if you want more notification targets in the future

src/main/kotlin/com/lambda/event/events/ModuleEvent.kt
IceTank: Having docs on top-level classes is good code documentation practice. When hovering over a class, it is best if you can get a brief description of what the class does instead of having to search for child methods or child classes and having to infer intent from those.

IceTank: All major Java and Kotlin libraries do this

IceTank: In fact, every Lambda class that is used in internal or external APIs should have this

beanbag44: Sure, having top-level class docs can be good, but most of this is unnecessary. Kotlin’s own advice for documentation is to:

```kotlin
// Avoid doing this:

/**
 * Returns the absolute value of the given number.
 * @param number The number to return the absolute value for.
 * @return The absolute value.
 */
fun abs(number: Int): Int { /*...*/ }

// Do this instead:

/**
 * Returns the absolute value of the given [number].
 */
fun abs(number: Int): Int { /*...*/ }
```

The ModuleEvent classes are incredibly clear on what they are designed for, and so don't require such bulky documentation. I dont typically ask for people to write docs in pr's because i dont really write them myself in the current stage of the clients development. This is due to things changing very frequently and docs quickly become outdated unless immediately updated, hence why i said you could just remove them. But, if you still wanna stick with them, just keep it concise. I dont see this class changing in the future.

src/main/kotlin/com/lambda/module/modules/client/ModuleNotifier.kt
beanbag44: i guess, maybe toasts? Not a big deal i suppose

IceTank:
I removed the event cancelation

PR #260, opened 2026-03-07
feat(BetterFirework): Improved takeoff reliability

IceTank:
Adds a setting to keep jumping to takeoff. It also works in water, so you don't have to spam your space bar to hit that one tick where you can use a firework. When enabled, it automatically holds down jump to get to the water surface and uses the firework when possible to take off.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
@@ -23,18 +23,19 @@ import com.lambda.config.applyEdits
 import com.lambda.config.settings.complex.Bind
 import com.lambda.config.settings.complex.KeybindSetting.Companion.onPress
 import com.lambda.context.SafeContext
+import com.lambda.event.events.MovementEvent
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/movement/BetterFirework.kt
```diff
@@ -117,7 +131,22 @@ object BetterFirework : Module(
 						connection.sendPacket(ClientCommandC2SPacket(player, ClientCommandC2SPacket.Mode.START_FALL_FLYING))
 					}
 					startFirework(invUse)
-					takeoffState = TakeoffState.None
+					takeoffState = if (jumpUntilTakeoff && player.isElytraEquipped) {
```
beanbag44, line 134: this seems to cause issues. When using the module with jumpUntilTakeoff, if you successfully takeoff and then land and jump within the jumpUntilTakeoffTimeout is over, it will continuously repeat the cycle as long as you keep jumping

beanbag44 (changes requested):

PR #262, opened 2026-03-08
feat(BlockPosSetting): Add button to set coordinates

IceTank:
Sets coordinates in the BlockPosSetting to the block you are currently looking at. Don't have to type out coordinates in chat anymore.

### DIFF
--- a/src/main/kotlin/com/lambda/config/Setting.kt
+++ b/src/main/kotlin/com/lambda/config/Setting.kt
@@ -230,13 +230,13 @@ class Setting<T : SettingCore<R>, R>(
 		buttonMenu = menu
 	}
 
-	fun trySetValue(newValue: R) {
+	fun trySetValue(newValue: R, logResponse: Boolean = true) {
 		if (newValue == value) {
-			ConfigCommand.info(notChangedMessage())
+			if (logResponse) ConfigCommand.info(notChangedMessage())
 		} else {
 			val previous = value
 			value = newValue
-			ConfigCommand.info(setMessage(previous, newValue))
+			if (logResponse) ConfigCommand.info(setMessage(previous, newValue))
 		}
 	}
 
--- a/src/main/kotlin/com/lambda/config/Setting.kt
+++ b/src/main/kotlin/com/lambda/config/Setting.kt
@@ -230,13 +230,13 @@ class Setting<T : SettingCore<R>, R>(
 		buttonMenu = menu
 	}
 
-	fun trySetValue(newValue: R, logResponse: Boolean = true) {
+	fun trySetValue(newValue: R) {
 		if (newValue == value) {
-			if (logResponse) ConfigCommand.info(notChangedMessage())
+			ConfigCommand.info(notChangedMessage())
 		} else {
 			val previous = value
 			value = newValue
-			if (logResponse) ConfigCommand.info(setMessage(previous, newValue))
+			ConfigCommand.info(setMessage(previous, newValue))
 		}
 	}
 
--- a/src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
+++ b/src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
@@ -26,11 +26,11 @@ import com.lambda.command.commands.ConfigCommand
 import com.lambda.config.Setting
 import com.lambda.config.SettingCore
 import com.lambda.gui.dsl.ImGuiBuilder
-import com.lambda.threading.runSafe
 import com.lambda.util.BlockUtils.blockPos
 import com.lambda.util.Communication.info
 import com.lambda.util.extension.CommandBuilder
 import com.lambda.util.world.raycast.RayCastUtils.blockResult
+import net.minecraft.client.MinecraftClient
 import net.minecraft.command.CommandRegistryAccess
 import net.minecraft.util.math.BlockPos
 
@@ -43,21 +43,16 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
 ) {
 	context(setting: Setting<*, BlockPos>)
 	override fun ImGuiBuilder.buildLayout() {
-		treeNode("Coordinates", id = setting.name) {
+		treeNode(setting.name, id = setting.name) {
 			inputVec3i(setting.name, value) { value = it.blockPos }
 		}
 		lambdaTooltip(setting.description)
 		sameLine()
 		button("Set") {
-			runSafe {
-				mc.crosshairTarget?.blockResult?.blockPos?.let {
-					setting.trySetValue(it, logResponse = false)
-					ConfigCommand.info("Coordinates updated")
-				} ?: let {
-					info("No block under cursor")
-					return@runSafe
-				}
-			}
+			MinecraftClient.getInstance().crosshairTarget?.blockResult?.blockPos?.let {
+				setting.trySetValue(it, logResponse = false)
+				ConfigCommand.info("Coordinates updated")
+			} ?: info("No block under crosshair")
 		}
 		lambdaTooltip("Set the coordinates to the block you are currently looking at")
 	}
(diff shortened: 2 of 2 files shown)


src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +42,36 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
...
+		}
+		lambdaTooltip(setting.description)
+		sameLine()
+		button("Set") {
+			runSafe {
+				mc.crosshairTarget?.blockResult?.blockPos?.let {
+					setting.trySetValue(it, logResponse = false)
+					ConfigCommand.info("Coordinates updated")
+				} ?: let {
+					info("No block under cursor")
```
beanbag44: should be "under crosshair"

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +42,36 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
...
+			inputVec3i(setting.name, value) { value = it.blockPos }
+		}
+		lambdaTooltip(setting.description)
+		sameLine()
+		button("Set") {
+			runSafe {
+				mc.crosshairTarget?.blockResult?.blockPos?.let {
+					setting.trySetValue(it, logResponse = false)
+					ConfigCommand.info("Coordinates updated")
+				} ?: let {
```
beanbag44: should use run instead of let as let passes a value into the block with none being used here

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +42,36 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
 	TypeToken.get(BlockPos::class.java).type
 ) {
 	context(setting: Setting<*, BlockPos>)
-    override fun ImGuiBuilder.buildLayout() {
-        inputVec3i(setting.name, value) { value = it.blockPos }
-        lambdaTooltip(setting.description)
-    }
+	override fun ImGuiBuilder.buildLayout() {
+		treeNode("Coordinates", id = setting.name) {
```
beanbag44: this doesnt display the name of the setting

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +42,36 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
...
+			inputVec3i(setting.name, value) { value = it.blockPos }
+		}
+		lambdaTooltip(setting.description)
+		sameLine()
+		button("Set") {
```
beanbag44: this displays even when the dropdown is closed

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +42,36 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
...
+			runSafe {
```
beanbag44: no need for runSafe here as although mc is included in SafeContext, the value of MinecraftClient.getInstance() is constant throughout the whole session and can be retreived through Lambda.kt

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +42,36 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
...
+				mc.crosshairTarget?.blockResult?.blockPos?.let {
+					setting.trySetValue(it, logResponse = false)
+					ConfigCommand.info("Coordinates updated")
+				} ?: let {
+					info("No block under cursor")
+					return@runSafe
```
beanbag44: no need for a return here as its the end of the block anyway

src/main/kotlin/com/lambda/config/Setting.kt
```diff
@@ -230,13 +230,13 @@ class Setting<T : SettingCore<R>, R>(
 		buttonMenu = menu
 	}
 
-	fun trySetValue(newValue: R) {
+	fun trySetValue(newValue: R, logResponse: Boolean = true) {
```
beanbag44: i dont think a logResponse parameter is required here. The button has the tooltip to explain what it does and the built-in log functionality offers an option to revert to the old value

beanbag44 (changes requested):
please test the code before publishing a pr

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
IceTank: That is on purpose. The idea is that you need to expand the tree node to see the coordinates. Otherwise coordinates will always show.

src/main/kotlin/com/lambda/config/Setting.kt
IceTank: I did it this way because the default response in chat shows coordinates of the setting you changed

beanbag44:
hm ok i see the point of the drop-down and different chat command now. The drop-down / button combination is still buggy though. When hovering over the drop-down, the highlight bar goes through the button, and i think for most people, they wont be clicking the button while streaming for example as its too close to opening the drop-down and leaking coordinates. Its prob best just to keep the set button in the drop-down. Maybe there could be an option once we add a Client module to obfuscate coordinates in settings unless clicked to edit them. Then the button could definitely be placed inside the drop-down, or even remove the drop-down as a whole.

beanbag44:
im still not a huge fan of overriding the default command print in chat as it completely removes the guarantee of an undo button. Im not sure the best way to do this, maybe adding an open command print function to SettingCore which appends the undo on the end would be a good way of keeping coherence between setting commands. For now i think just stick with the standard command and we can come up with a more flexible system in the future

IceTank:
I mean, when you change values in the GUI, there is no undo button? Why should there be an undo button for this specific setting?

IceTank:
Resolved merge conflict

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +41,31 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
 	TypeToken.get(BlockPos::class.java).type
 ) {
 	context(setting: Setting<*, BlockPos>)
-    override fun ImGuiBuilder.buildLayout() {
-        inputVec3i(setting.name, value) { value = it.blockPos }
-        lambdaTooltip(setting.description)
-    }
+	override fun ImGuiBuilder.buildLayout() {
+		button("Set") {
+			MinecraftClient.getInstance().crosshairTarget?.blockResult?.blockPos?.let {
```
beanbag44: mc from Lambda.kt over MinecraftClient.getInstance()

src/main/kotlin/com/lambda/config/settings/complex/BlockPosSetting.kt
```diff
@@ -38,21 +41,31 @@ class BlockPosSetting(defaultValue: BlockPos) : SettingCore<BlockPos>(
...
+				setting.trySetValue(it, logResponse = false)
```
beanbag44: why not set the settings value as it does below? value = it. That would avoid the whole overriding the command print thing in chat

beanbag44 (changes requested):

src/main/kotlin/com/lambda/config/Setting.kt
beanbag44: theres no need for the changes here now as the value is set directly

beanbag44 (changes requested):
good, just one small thing

PR #263, opened 2026-03-08
feat(AutoVillagerCycle): Add module to automatically cycle librarians for book enchantments

IceTank:
This module automatically cycles librarians by continuously breaking and replacing lecterns until a specific enchanted book is found.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
+++ b/src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
@@ -37,6 +37,7 @@ import com.lambda.task.Task
 import com.lambda.task.tasks.BuildTask.Companion.build
 import com.lambda.threading.runSafeAutomated
 import com.lambda.util.BlockUtils.blockState
+import com.lambda.util.BlockUtils.isEmpty
 import com.lambda.util.Communication.info
 import com.lambda.util.Communication.logError
 import com.lambda.util.NamedEnum
@@ -56,7 +57,7 @@ import net.minecraft.util.math.BlockPos
 
 
 object AutoVillagerCycle : Module(
-	name = "AutoVillagerCycler",
+	name = "AutoVillagerCycle",
 	description = "Automatically cycles librarian villagers with lecterns until a desired enchanted book is found",
 	tag = ModuleTag.PLAYER
 ) {
@@ -74,19 +75,19 @@ object AutoVillagerCycle : Module(
 	private val searchRange by setting("Search Range", 5.0, 1.0..10.0, 0.5, "Range to search for nearby villagers", " blocks").group(Group.General)
 	private val startCyclingBind by setting("Start Cycling", Bind.EMPTY, "Press to start/stop cycling").group(Group.General)
 		.onPress {
-			if (cycleState != CycleState.IDLE) {
+			if (cycleState != CycleState.Idle) {
 				info("Stopped villager cycling.")
-				switchState(CycleState.IDLE)
+				switchState(CycleState.Idle)
 			} else {
 				info("Started villager cycling.")
 				buildTask?.cancel()
 				buildTask = null
-				switchState(CycleState.PLACE_LECTERN)
+				switchState(CycleState.PlaceLectern)
 			}
 		}
 	private val desiredEnchantments by setting("Desired Enchantments", emptySet(), allEnchantments).group(Group.Enchantments)
 
-	private var cycleState = CycleState.IDLE
+	private var cycleState = CycleState.Idle
 	private var tickCounter = 0
 
 	private var buildTask: Task<*>? = null
@@ -97,12 +98,12 @@ object AutoVillagerCycle : Module(
 		onEnable {
 			allEnchantments.clear()
 			allEnchantments.addAll(getEnchantmentList())
-			cycleState = CycleState.IDLE
+			cycleState = CycleState.Idle
 			tickCounter = 0
 		}
 
 		onDisable {
-			cycleState = CycleState.IDLE
+			cycleState = CycleState.Idle
 			tickCounter = 0
 			buildTask?.cancel()
 			buildTask = null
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+import net.minecraft.network.packet.s2c.play.SetTradeOffersS2CPacket
+import net.minecraft.registry.RegistryKeys
+import net.minecraft.sound.SoundEvents
+import net.minecraft.util.Hand
+import net.minecraft.util.hit.EntityHitResult
+import net.minecraft.util.math.BlockPos
+
+
+object AutoVillagerCycle : Module(
+	name = "AutoVillagerCycler",
```
beanbag44: module name should match the object name

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+
+	private fun switchState(newState: CycleState) {
+		if (cycleState != newState) {
+			tickCounter = 0
+		}
+		cycleState = newState
+	}
+
+	private enum class CycleState {
+		IDLE,
```
beanbag44: camel-case

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	private val allEnchantments = ArrayList<String>()
+
+	private val lecternPos by setting("Lectern Pos", BlockPos.ORIGIN, "Position where the lectern should be placed/broken").group(Group.General)
+	private val logFoundBooks by setting("Log Found Books", true, "Log all enchanted books found during cycling").group(Group.General)
+	private val interactDelay by setting("Interact Delay", 20, 1..40, 1, "Ticks to wait before interacting with the villager", " ticks").group(Group.General)
+	private val breakDelay by setting("Break Delay", 5, 1..20, 1, "Ticks to wait after breaking the lectern", " ticks").group(Group.General)
+	private val searchRange by setting("Search Range", 5.0, 1.0..10.0, 0.5, "Range to search for nearby villagers", " blocks").group(Group.General)
+	private val startCyclingBind by setting("Start Cycling", Bind.EMPTY, "Press to start/stop cycling").group(Group.General)
```
beanbag44, line 79: why not have the module itself start the cycling?

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+
+		if (lecternPos == BlockPos.ORIGIN) {
+			logError("Lectern position is not set!")
+			switchState(CycleState.IDLE)
+			return
+		}
+
+		val state = blockState(lecternPos)
+
+		if (!state.isAir) {
```
beanbag44: this could probably be replaced with isEmpty to account for if someone is attempting to place the lectern in a waterlogged position

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+			logError("No villager found nearby!")
+			switchState(CycleState.IDLE)
+			return
+		}
+
+		runSafeAutomated {
+			lookAtEntity(villager)?.let {
+				rotationRequest {
+					rotation(it.rotation)
+				}.submit()
```
beanbag44: theres no check here to see if the request is done

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+	private fun SafeContext.handleBreakLectern() {
+		if (player.currentScreenHandler != player.playerScreenHandler) {
+			player.closeHandledScreen()
+		}
+
+		if (tickCounter < breakDelay) return
+
+		val state = blockState(lecternPos)
+
+		if (!state.isAir) {
```
beanbag44: isEmpty

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+	private fun SafeContext.handleOpenVillager() {
+		if (tickCounter < interactDelay) return
+
+		if (buildTask?.state == Task.State.Running) {
+			return
+		}
+
+		// Verify lectern is still present
+		val state = blockState(lecternPos)
+		if (state.isAir) {
```
beanbag44: isEmpty

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,301 @@
...
+		}
+	private val desiredEnchantments by setting("Desired Enchantments", emptySet(), allEnchantments).group(Group.Enchantments)
+
+	private var cycleState = CycleState.IDLE
+	private var tickCounter = 0
+
+	private var buildTask: Task<*>? = null
+
+	init {
+		setDefaultAutomationConfig()
```
beanbag44: displaying the full automation config doesnt cause any issues, but there are likely a lot of unused settings within here. My guess is that we only really need to have the interact, rotation and break configs. Maybe im missing some but you get the gist

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
IceTank: Idk I like it that way. But I can change it. I generally have a start automation key that I use to start different modules. For instance when I want to autocraft I start it with f and disable the module when I don't have to craft anything. With the villager cyler I would put it on f as well and disable the module when I want to use a different automation module.

beanbag44: fair, that makes a lot of sense. We can keep it this way then

src/main/kotlin/com/lambda/module/modules/player/AutoVillagerCycle.kt
```diff
@@ -0,0 +1,304 @@
...
+			switchState(CycleState.Idle)
+			return
+		}
+
+		runSafeAutomated {
+			lookAtEntity(villager)?.let {
+				val done = rotationRequest {
+					rotation(it.rotation)
+				}.submit().done
+				if (done) {
```
beanbag44, line 264: the tickCounter is set back to 0 even if the rotation request fails. That could probably result in a constant loop if the user doesnt have instant rotations as it would attempt to rotate. Not fast enough to complete it the first tick, then fail the interact, delay would let the rotation go back to where it was, and then so on and so on

beanbag44 (changes requested):
one small thing

PR #266, opened 2026-03-12
Fix - KillAura cannot target projectiles and other non-LivingEntities.

TwinkNet:
Previously lambda's killaura was unable to target and hit projectiles, like shulker bullets, despite having the ability to select them as a target. This is because `Targeting.target()` would only ever give you a `LivingEntity`, below is the original code for `Targeting.target()`:

```kt
fun target(): LivingEntity? = runSafe {
    return@runSafe fastEntitySearch<LivingEntity>(targetingRange) {
    // 5 lines not shown...
```

This is changed in my pull to accept a `T : Entity` generic during the call, so that in `KillAura.kt`:
```kt
val target: LivingEntity?
    get() = targeting.target()
```
becomes
```kt
val target: Entity?
    get() = targeting.target<Entity>()
```
This allows more flexibility in the use of the targeting system, and allows `CrystalAura.kt` to remain unaffected by this change, it'll still only target for a LivingEntity:
```kt
private fun SafeContext.tick() {
    // Update the target
    currentTarget = targeting.target<LivingEntity>()
```

some minor modifications to `Targeting.validate()` and `Targeting.Priority` were made to accommodate a generic `Entity` object

### DIFF
--- a/src/main/kotlin/com/lambda/config/groups/Targeting.kt
+++ b/src/main/kotlin/com/lambda/config/groups/Targeting.kt
@@ -105,7 +105,7 @@ abstract class Targeting(
          * Validates whether a given entity is targetable for combat based on the field of view limit and other settings.
          *
          * @param player The [ClientPlayerEntity] performing the targeting.
-         * @param entity The [LivingEntity] being evaluated.
+         * @param entity The [Entity] being evaluated.
          * @return `true` if the entity is valid for targeting, `false` otherwise.
          */
         override fun validate(player: ClientPlayerEntity, entity: Entity): Boolean {
--- a/src/main/kotlin/com/lambda/config/groups/Targeting.kt
+++ b/src/main/kotlin/com/lambda/config/groups/Targeting.kt
@@ -149,7 +149,7 @@ abstract class Targeting(
 
         /**
          * Prioritizes entities based on their health.
-         * Entites that aren't an instanceof LivingEntity will be treated as if they have Double.MAX_VALUE health,
+         * Entities that aren't an instanceof LivingEntity will be treated as if they have Double.MAX_VALUE health,
          * therefore having least priority
          */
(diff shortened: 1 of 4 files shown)


emyfops:
There is no need to pass a type to the entity to the `target` function. Just return `Entity` since it's not used anywhere else.

emyfops:
> There is no need to pass a type to the entity to the `target` function. Just return `Entity` since it's not used anywhere else.

I get the idea but I'm not a fan of it just to target non-living entities. Most, if not all modules will have to use the targetting class without the delegate and it'll make a mess.

emyfops:
> > There is no need to pass a type to the entity to the `target` function. Just return `Entity` since it's not used anywhere else.
> 
> I get the idea but I'm not a fan of it just to target non-living entities. Most, if not all modules will have to use the targetting class without the delegate and it'll make a mess.

Ok maybe it is a good idea after all, I approve of this.

src/main/kotlin/com/lambda/config/groups/Targeting.kt
```diff
@@ -140,19 +138,21 @@ abstract class Targeting(
...
          */
         Distance({ player.pos distSq it.pos }),
 
         /**
          * Prioritizes entities based on their health.
+         * Entites that aren't an instanceof LivingEntity will be treated as if they are at 0 health,
+         * therefore having least priority
          */
-        Health({ it.fullHealth }),
+        Health({ (it as? LivingEntity)?.fullHealth ?: 0.0 }),
```
beanbag44: Everything looks good, but won't non-living entities be selected over living entities because it returns 0.0 and sorts by minByOrNull?

beanbag44 (changes requested):

src/main/kotlin/com/lambda/config/groups/Targeting.kt
TwinkNet: Ah you're correct, I'll fix this.

PR #267, opened 2026-03-12
Fix: use distance from boundbox edge for targeting distance

TwinkNet:
Currently Lambda uses the `Entity.pos` to calculate distance from the player, which results in inconsistent targeting when fighting entities with large hitboxes, such as big slimes, ender dragons, and withers. I keep my targeting distance set to 3 blocks and there are certain mobs that it just won't hit unless i'm kissing them.

This PR adds a helper function for calculating the distance from a `Vec3d` to the closest side of a `Box`, and uses that function in `WorldUtils.internalGetFastEntities()` to calculate distance.

Additionally, `fastEntitySearch()` now accepts a `FastVector` rather than a `BlockPos`, as I found that this helps achieve a bit more consistency in targeting when standing on the edge of a block.


src/main/kotlin/com/lambda/util/world/WorldUtils.kt
```diff
@@ -79,6 +81,19 @@ object WorldUtils {
         }
     }
 
+    /**
+     * Returns the distance squared to the closest side of a [Box] to the specified [point]
+     */
+    fun getDistSqToClosestSideOfBoundBox(point: Vec3d, box: Box): Double {
```
emyfops: This should be in the Vectors.kt file. The naming should be simplified and let's opt-in for an extension on Vec3d.

Ex: 
fun Vec3d.distanceToSide(box: Box) =

emyfops (changes requested):

src/main/kotlin/com/lambda/util/world/WorldUtils.kt
TwinkNet: Thanks for the tips, the changes have been made. The code looks much cleaner now.

cattyngmd:
you need to account for AABB epsilon when calculating distance to closest AABB edge

src/main/kotlin/com/lambda/util/math/Vectors.kt
```diff
@@ -202,6 +203,19 @@ infix fun Entity.distSq(other: Vec3d): Double = pos distSq other
 infix fun Entity.distSq(other: Vec3i): Int = blockPos distSq other
 infix fun Entity.distSq(other: Entity): Double = squaredDistanceTo(other)
 
+/**
+ * Returns the distance squared to the closest side of a [Box]
+ */
+fun Vec3d.distanceToSide(box: Box, epsilon: Double = 1e-7): Double {
```
beanbag44: this seems to return a squared distance. Maybe a better name could be distToClosestPointSq? Or without the Sq if we square root it. Maybe another version of the function could be made for FastVector too to avoid the transformation in world utils

beanbag44 (changes requested):

beanbag44 (changes requested):
Sorry its taken me a while to get around to reviewing pr's. This changes the internalGetFastEntities function, but the change should also be applied to internalGetEntities. I still think a second function for getting the distance to the edge of a bounding box using FastVector would be nice to avoid the extra conversion to Vec3d in WorldUtils.kt. They could also be changed to infix functions like how Vec3d.distSq does it.

TwinkNet:
Yeah I got busy with another project and haven't been able to finish this. I've converted this to a draft for now.

emyfops:
So the problem here is that `FastVector` does not have floating point capabilities. It's been so long I'm having a hard time changing this.

Basically, functions that check for the distance between a vector and an entity shall no longer take in a `FastVector` as to not lose accuracy. I did not think about checking the distance to the bounding box in 2024 :sweat:

emyfops:
@TwinkNet Could you confirm whether these were the changes you were expecting and how to verify if it works correctly ?

beanbag44:
gonna close this pr for now as its been inactive for a while

PR #269, opened 2026-03-14
feat(Freecam): Add Exclude Rotation mode to freecam

IceTank:
This mode lets you configure what rotations are applied to the player when rotation mode in freecam is enabled.
This is useful for when you are in freecam and want to be able to change the view direction of the player without exiting freecam. Possible use cases are elytra flying while in freecam and changing the player's yaw to the freecam yaw while flying.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
+++ b/src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
@@ -26,6 +26,7 @@ import com.lambda.event.events.PlayerEvent
 import com.lambda.event.events.RenderEvent
 import com.lambda.event.events.TickEvent
 import com.lambda.event.listener.SafeListener.Companion.listen
+import com.lambda.interaction.managers.rotating.IRotationRequest
 import com.lambda.interaction.managers.rotating.IRotationRequest.Companion.rotationRequest
 import com.lambda.interaction.managers.rotating.Rotation
 import com.lambda.interaction.managers.rotating.RotationMode
@@ -77,6 +78,7 @@ object Freecam : Module(
 	private val sprint by setting("Sprint Multiplier", 3.0, 0.1..10.0, 0.1, description = "Set below 1.0 to fly slower on sprint.") { mode == Mode.Free }
 	private val reach by setting("Reach", 10.0, 1.0..100.0, 1.0, "Freecam reach distance")
 	private val rotateMode by setting("Rotate Mode", FreecamRotationMode.None, "Rotation mode").onValueChange { _, it -> if (it == FreecamRotationMode.LookAtTarget) mc.crosshairTarget = BlockHitResult.createMissed(Vec3d.ZERO, Direction.UP, BlockPos.ORIGIN) }
+	private val excludeRotation by setting("Exclude Rotation", FreecamRotationExclusion.None, description = "Exclude certain rotation changes from rotate mode") { rotateMode != FreecamRotationMode.None }
 	private val relative by setting("Relative", false, "Moves freecam relative to player position") { mode == Mode.Free }.onValueChange { _, it -> if (it) lastPlayerPosition = player.pos }
 	private val keepYLevel by setting("Keep Y Level", false, "Don't change the camera y-level on player movement") { mode == Mode.Free && relative }
 
@@ -141,14 +143,6 @@ object Freecam : Module(
 
 		onDisable {
 			mc.options.perspective = lastPerspective
-		}
-
-		listen<TickEvent.Pre> {
-			when (rotateMode) {
-				FreecamRotationMode.None -> return@listen
-				FreecamRotationMode.KeepRotation -> rotationRequest { rotation(rotation) }.submit()
-				FreecamRotationMode.LookAtTarget -> mc.crosshairTarget?.let { rotationRequest { rotation(lookAt(it.pos)) }.submit() }
-			}
 		}
 
 		listen<PlayerEvent.ChangeLookDirection> {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -227,6 +221,33 @@ object Freecam : Module(
...
+					buildRotationTo(lookAt(it.pos))
+				}.submit() }
+			}
+		}
+	}
+
+	context(rotationContext: IRotationRequest.RotationRequestBuilder)
+	private fun buildRotationTo(rotation: Rotation) {
+		when (excludeRotation) {
+			FreecamRotationExclusion.Pitch -> {
```
beanbag44: can inline these to avoid the brackets

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -227,6 +221,33 @@ object Freecam : Module(
 			mc.crosshairTarget = rotation.rayCast(reach, lerpPos).orMiss // Can't be null (otherwise mc will spam "Null returned as 'hitResult', this shouldn't happen!")
 			mc.crosshairTarget?.let { if (it.type != HitResult.Type.MISS) event.cancel() }
 		}
+
+		listen<TickEvent.Pre> {
+			when (rotateMode) {
+				FreecamRotationMode.None -> return@listen
```
beanbag44: can change the None case to null and remove the submit calls for the other outcomes. Then you can move the .submit to the end of the when block with ?.submit(). Or the submit calls could be moved into the buildRotationTo function

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -227,6 +221,33 @@ object Freecam : Module(
...
+				}.submit()
+				FreecamRotationMode.LookAtTarget -> mc.crosshairTarget?.let { rotationRequest {
+					buildRotationTo(lookAt(it.pos))
+				}.submit() }
+			}
+		}
+	}
+
+	context(rotationContext: IRotationRequest.RotationRequestBuilder)
+	private fun buildRotationTo(rotation: Rotation) {
```
beanbag44: rather than passing a rotation request context, you could just have this function create the request and return it. rotationRequest { *when block with builder calls* }

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -227,6 +221,33 @@ object Freecam : Module(
...
+			FreecamRotationExclusion.Yaw -> {
+				rotationContext.pitch(rotation.pitch)
+			}
+			else -> {
+				rotationContext.rotation(rotation)
+			}
+		}
 	}
 
 	private enum class FreecamRotationMode(override val displayName: String, override val description: String) : NamedEnum, Describable {
```
beanbag44: can remove the Freecam prefix for these enums. Theyre private so theres no need to preface what theyre for. Also while you're cleaning things up, might as well move these enums below the function at the bottom and remove the extra white space under the followTrackPlayer setting

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -77,6 +78,7 @@ object Freecam : Module(
 	private val sprint by setting("Sprint Multiplier", 3.0, 0.1..10.0, 0.1, description = "Set below 1.0 to fly slower on sprint.") { mode == Mode.Free }
 	private val reach by setting("Reach", 10.0, 1.0..100.0, 1.0, "Freecam reach distance")
 	private val rotateMode by setting("Rotate Mode", FreecamRotationMode.None, "Rotation mode").onValueChange { _, it -> if (it == FreecamRotationMode.LookAtTarget) mc.crosshairTarget = BlockHitResult.createMissed(Vec3d.ZERO, Direction.UP, BlockPos.ORIGIN) }
```
beanbag44: onValueChange can be placed on the next line. Also you sure we need this onValueChange?

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -77,6 +78,7 @@ object Freecam : Module(
...
+	private val excludeRotation by setting("Exclude Rotation", FreecamRotationExclusion.None, description = "Exclude certain rotation changes from rotate mode") { rotateMode != FreecamRotationMode.None }
 	private val relative by setting("Relative", false, "Moves freecam relative to player position") { mode == Mode.Free }.onValueChange { _, it -> if (it) lastPlayerPosition = player.pos }
```
beanbag44: onValueChanged on the next line again

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
IceTank: Yes. Otherwise, the relative delta calculation would mess up and teleport the freecam position somewhere else.

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
IceTank: Can't. RotationMode is already used when configuring the rotation config. The names would conflict.

beanbag44: The other enum can. Looking at the rotation configs enum, you can move that into the RotationConfig interface at the bottom and then reference it directly through the interface when setting the rotation config. The other configs do this, the rotation config is just a bit old and doesnt do that currently. That would allow you to remove the Freecam suffix

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
beanbag44: i mean the crosshair setting onValueChange

beanbag44: if (it == FreecamRotationMode.LookAtTarget) mc.crosshairTarget...

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -227,17 +225,30 @@ object Freecam : Module(
...
 			mc.crosshairTarget?.let { if (it.type != HitResult.Type.MISS) event.cancel() }
 		}
-	}
 
-	private enum class FreecamRotationMode(override val displayName: String, override val description: String) : NamedEnum, Describable {
-		None("None", "No rotation changes"),
-		LookAtTarget("Look At Target", "Look at the block or entity under your crosshair"),
-		KeepRotation("Keep Rotation", "Look in the same direction as the camera");
+		listen<TickEvent.Pre> {
+			runSafeAutomated {
```
beanbag44, line 230: you dont need this runSafeAutomated

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -247,4 +258,21 @@ object Freecam : Module(
 		val players = world.players.filter { it !is ClientPlayerEntity }
 		return players.minByOrNull { it.eyePos.squaredDistanceTo(position) }
 	}
+
+	private enum class RotationModeFreecam(override val displayName: String, override val description: String) : NamedEnum, Describable {
```
beanbag44, line 262: this is still suffixed with Freecam

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -247,4 +258,21 @@ object Freecam : Module(
...
+		None("None", "No rotation changes"),
+		LookAtTarget("Look At Target", "Look at the block or entity under your crosshair"),
+		KeepRotation("Keep Rotation", "Look in the same direction as the camera");
+	}
+
+	private enum class FreecamRotationExclusion(override val displayName: String, override val description: String) : NamedEnum, Describable {
```
beanbag44, line 268: this is still prefixed with Freecam

src/main/kotlin/com/lambda/module/modules/player/Freecam.kt
```diff
@@ -227,17 +225,30 @@ object Freecam : Module(
...
-	private enum class FreecamRotationMode(override val displayName: String, override val description: String) : NamedEnum, Describable {
-		None("None", "No rotation changes"),
-		LookAtTarget("Look At Target", "Look at the block or entity under your crosshair"),
-		KeepRotation("Keep Rotation", "Look in the same direction as the camera");
+		listen<TickEvent.Pre> {
+			runSafeAutomated {
+				when (rotateMode) {
+					RotationModeFreecam.None -> null
+					RotationModeFreecam.KeepRotation -> rotationRequest {
+						buildRotationTo(rotation)
```
beanbag44, line 234: have you tested this?

beanbag44 (changes requested):

PR #270, opened 2026-03-15
feat(KillAura): Add Target Named Entities to targeting setting

IceTank:
This setting prevents kill aura from targeting entities that are named by default


src/main/kotlin/com/lambda/config/groups/Targeting.kt
```diff
@@ -103,6 +105,11 @@ abstract class Targeting(
          */
         val priority by c.setting("${prefix}Priority", Priority.Distance, visibility = visibility).group(*baseGroup).index()
 
+	    /**
+	     * Whether to target named entities (e.g., players with custom names). Configurable with default set to `true`.
```
beanbag44: comment still says e.g., players with custom names

beanbag44 (changes requested):

beanbag44:
isDead is not present on Entity. You still need to do the cast to LivingEntity

IceTank:
> isDead is not present on Entity. You still need to do the cast to LivingEntity

That change came from upstream when resolving a merge conflict. I might have messed up but I didnt write the line I just merged them together.

PR #272, opened 2026-03-18
Feature: Auto Portal Builder

beanbag44:
Automatically places, lights, and walks into a nether portal using the given configuration.

Once the module is enabled, you can bind the previewPlace setting which once pressed shows you a preview of where the portal will be placed. Once released, the module will place the portal and do other actions if configured for such.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
+++ b/src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
@@ -22,6 +22,7 @@ import com.lambda.config.AutomationConfig.Companion.setDefaultAutomationConfig
 import com.lambda.config.applyEdits
 import com.lambda.config.groups.WorldLineSettings
 import com.lambda.config.settings.complex.Bind
+import com.lambda.config.settings.complex.KeybindSetting.Companion.onPress
 import com.lambda.config.settings.complex.KeybindSetting.Companion.onRelease
 import com.lambda.context.SafeContext
 import com.lambda.event.events.TickEvent
@@ -77,7 +78,9 @@ object AutoPortal : Module(
 	}
 
 	private val previewPlace by setting("Preview Place", Bind.EMPTY, "The keybind to preview the portal placement and subsequentially place the portal").group(Group.General)
+		.onPress { preview = true }
 		.onRelease {
+			preview = false
 			buildTask?.cancel()
 			val posStateMap =
 				obiPositions.associateWith {
@@ -120,6 +123,7 @@ object AutoPortal : Module(
 		}
 	}
 
+	private var preview = false
 	private var buildTask: Task<*>? = null
 
 	init {
@@ -137,8 +141,7 @@ object AutoPortal : Module(
 		}
 
 		immediateRenderer("AutoPortal Immediate Renderer", { depthTest }) { safeContext ->
-			if (!renders) return@immediateRenderer
-			if (!previewPlace.isSatisfied()) return@immediateRenderer
+			if (!renders || !preview) return@immediateRenderer
 			with (safeContext) {
 				val obiColor = blockColor(Blocks.OBSIDIAN.defaultState, BlockPos.ORIGIN)
 				obiPositions
--- a/src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
+++ b/src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
@@ -1,0 +1,321 @@
+/*
+ * Copyright 2026 Lambda
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.lambda.module.modules.player
+
+import baritone.api.pathing.goals.GoalXZ
+import com.lambda.config.AutomationConfig.Companion.setDefaultAutomationConfig
+import com.lambda.config.applyEdits
+import com.lambda.config.groups.WorldLineSettings
+import com.lambda.config.settings.complex.Bind
+import com.lambda.config.settings.complex.KeybindSetting.Companion.onRelease
+import com.lambda.context.SafeContext
+import com.lambda.event.events.TickEvent
+import com.lambda.event.listener.SafeListener.Companion.listen
+import com.lambda.graphics.mc.renderer.ImmediateRenderer.Companion.immediateRenderer
+import com.lambda.graphics.util.DirectionMask.buildSideMesh
+import com.lambda.interaction.BaritoneManager
+import com.lambda.interaction.construction.verify.TargetState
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+			buildTask = posStateMap
+				.build()
+				.thenOrNull {
+					if (light) LightTask(currAnchorPos.up(), walkIn)
+					else null
+				}
+				.finally {
+					buildTask = null
+				}
+				.run()
```
IceTank, line 104: Probably better to extra this logic into a method

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+		}
+	private val corners by setting("Corners", false).group(Group.General)
+	private val light by setting("Light", true, "Attempts to automatically light the portal after building").group(Group.General)
+	private val walkIn by setting("Walk In", true, "Automatically paths into the portal with baritone") { light }.group(Group.General)
```
IceTank, line 108: Why hide when light is off?

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+	private val inventory by setting("Inventory", true, "Allows access to the players inventory when retrieving a flint and steel for lighting the portal").group(Group.General)
+	private val forwardOffset by setting("Forward Offset", 3, 0..10).group(Group.General)
+	private val sidewaysOffset by setting("Sideways Offset", 0, -5..5).group(Group.General)
+	private val yOffset by setting("Y Offset", 0, -5..5).group(Group.General)
+	private val lockToGround by setting("Lock To Ground", true).group(Group.General)
+	private val allowUpwardShift by setting("Allow Upward Shift", true).group(Group.General)
```
IceTank: Missing setting description on allowUpwardShift

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+									prevAnchorPos.subtract(currAnchorPos).vec3d
+								)
+								box.offset(offset)
+							} else box
+						}
+						Pair(it, box)
+					}
+					.forEach { posAndBox ->
+						box(posAndBox.second, outlineConfig) {
+							colors(obiColor.setAlpha(0.3), obiColor)
```
IceTank: Would be nice to be able to change the alpha value of the render

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+		private val originObiPositionsWithCorners = getOriginObiPositions(true)
+		private val originPortalPositions = getOriginPortalPositions()
+
+		context(safeContext: SafeContext)
+		fun tick() =
+			with(safeContext) {
+				if (!previewPlace.isSatisfied()) return@with
+				val offsetDir = when (val playerHorizontal = player.horizontalFacing) {
+					Direction.UP,
+					Direction.DOWN -> Direction.EAST
```
IceTank: playerHorizontal is always one of the 4 horizontal directions never UP or DOWN. So this when is redundant.

IceTank: You would need to make a separate check to handle up and down

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+
+				val baseAnchorPos = player.blockPos
+					.offset(offsetDir, forwardOffset)
+					.offset(offsetDir.rotateYClockwise(), sidewaysOffset)
+
+				val lockedAnchorPos = if (lockToGround) run {
+					var scanPos = baseAnchorPos
+					if (blockState(scanPos).isNotEmpty) {
+						if (allowUpwardShift) {
+							while (blockState(scanPos).isNotEmpty && scanPos.y < 320) {
```
IceTank: Why 320 specifically? Max build height is 319, and I am pretty sure you won't be able to build a portal if the lowest block is already at build height. Maybe just fail if the y level is to heigh?

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+								scanPos = scanPos.up()
+							}
+						}
+						if (!allowUpwardShift || scanPos.y >= 320) {
```
IceTank: Same as above comment

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
```diff
@@ -0,0 +1,355 @@
...
+								pos,
+								false,
+								false
+							),
+							sequence
+						)
+					}
+					swapPacket()
+					if (walkIn) {
+						BaritoneManager.setGoalAndPath(GoalXZ(pos.x, pos.z))
```
IceTank: Should be a GoalBlock. Preferably a
```suggestion
BaritoneManager.setGoalAndPath(GoalComposite(GoalBlock(...), GoalBlock(...)))
```
so baritone walks into either portal block.

IceTank (changes requested):

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
beanbag44: how so? Having it here makes it clear and concise what it does. Putting it in another function just makes it more confusing when navigating how the module works

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
beanbag44: because theres not much point walking into the portal when its not lit. Helps keep the setting bloat down by assuming what most people would want

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
beanbag44: didnt bother adding it because lockToGround implies that its locking down to the ground. allowUpwardShift would imply youre letting it rize to the top of some blocks.

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
beanbag44: was trying to keep down on the setting bloat and wasnt sure if i should add this. Seems very niche and i doubt people would bother changing it most the time but ig i can

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
beanbag44: true

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
beanbag44: i based the y levels off where it would be technically possible to place the base blocks of the portal with a support. I chose 320 because i need a way to see if the upward shift was successful, and as 319 could be a valid placement, there would be no way to know if it failed or not.

src/main/kotlin/com/lambda/module/modules/player/AutoPortal.kt
beanbag44: ah yeah. I tried goal block but i assumed at the time it only walked near that position. Turns out i was telling it to walk into the anchor pos obsidian

Issue #254, opened 2026-03-07
Bug: ContainerPreview does not display shulker name


IceTank:
### What happened?

Instead of the shulker name the container preview displays no name on previews

### What should have happened?

ContainerPreview shows the container name in the preview

### Version

1.21.11 latest dev build

PR #275, opened 2026-03-21
fix(ContainerPreview): Fix invisible text on preview

IceTank:
This pr changes the luminance method. It replaces the magic number values with the Minecraft internal color values. The used magic values had zero alpha components and thus appeared invisible.
This PR also changes the luminance check on the text color to make the text in the preview windows easier to read.

Luminance comparison
Before:
image
image

After:
image
image


IceTank:
fixes #254

PR #276, opened 2026-03-22
Feature: LightLevels module

beanbag44:
adds renders to show when the light level gets below a given threshold on spawnable blocks. Has two modes; chunked and radius. Chunked will scan the entire loaded render distance, and radius will scan a smaller radius around the player


src/main/kotlin/com/lambda/module/modules/render/LightLevels.kt
```diff
@@ -0,0 +1,149 @@
...
+		Line("Line")
+	}
+
+	private val mode: Mode by setting("Mode", Mode.Chunked)
+		.onValueChange { _, _ -> chunkedRenderer.clear(); refreshChunkedRenderer(this) }
+	private val minLightLevel by setting("Min Light Level", 0, 0..15).onValueChange(::refreshChunkedRenderer)
+	private val renderMode by setting("Render Mode", RenderMode.Square).onValueChange(::refreshChunkedRenderer)
+	private val color by setting("Color", Color.RED).onValueChange(::refreshChunkedRenderer)
+	private val size by setting("Size", 14, 1..16).onValueChange(::refreshChunkedRenderer)
+	private val fill by setting("Fill", false) { renderMode == RenderMode.Square }.group(Group.Fill).onValueChange(::refreshChunkedRenderer)
```
IceTank, line 61: Either put `.onValueChange` in a newline or not. You are doing both.

src/main/kotlin/com/lambda/module/modules/render/LightLevels.kt
```diff
@@ -0,0 +1,149 @@
...
+	private fun refreshChunkedRenderer(ctx: SafeContext, from: Any? = null, to: Any? = null) {
+		if (mode == Mode.Chunked) chunkedRenderer.rebuild()
+	}
+
+	private enum class Mode {
+		Chunked,
+		Radius
+	}
+
+	private enum class RenderMode {
```
IceTank, line 147: Maybe extend `NamedEnum` and `Describable` ?

src/main/kotlin/com/lambda/module/modules/render/LightLevels.kt
```diff
@@ -0,0 +1,149 @@
...
+	private val chunkedRenderer = chunkedRenderer("LightLevels Chunked Renderer", { depthTest }) { _, pos ->
+		if (mode != Mode.Chunked) return@chunkedRenderer
+		runSafe { buildRender(pos.toBlockPos()) }
+	}
+
+	init {
+		tickedRenderer("LightLevels Ticked Renderer", { depthTest }) {
+			if (mode != Mode.Radius) return@tickedRenderer
+			runSafe {
+				val playerPos = mc.gameRenderer.camera.pos.flooredBlockPos
```
IceTank: Why flooredBlockPos? Why not more accurate with double values?

src/main/java/com/lambda/mixin/world/ClientChunkManagerMixin.java
```diff
@@ -19,70 +19,50 @@
...
-//            WorldChunk oldChunk,
-//            ChunkPos chunkPos
-//    ) {
-//        if (!clientChunkMap.isInRadius(chunkPos.x, chunkPos.z)) {
-//            EventFlow.post(new WorldEvent.ChunkEvent.Unload(this.world, oldChunk));
-//        }
-//    }
+    @Inject(method = "onLightUpdate", at = @At("RETURN"))
+    private void injectOnLightUpdate(LightType type, ChunkSectionPos pos, CallbackInfo ci) {
+        if (LightLevels.INSTANCE.isEnabled()) {
```
IceTank, line 64: Why not make an event?

IceTank (changes requested):
I eat sand

src/main/kotlin/com/lambda/module/modules/render/LightLevels.kt
beanbag44: In this case is makes more sense to have them on the same line as it is the same onValueChange repeated for most of the render settings and is small. Unique onValueChange's should be put on a new line as they offer new logic which should be more visible and readable. This is why the onValueChange is on a new line for the mode setting

src/main/kotlin/com/lambda/module/modules/render/LightLevels.kt
beanbag44: it makes no difference. If you add 8 to 1.9999 and then floor it, it gives you the same value as 8 + 1. This would only be somewhat worth it if our range settings werent whole integers, but theres not much point in that

src/main/kotlin/com/lambda/module/modules/render/LightLevels.kt
beanbag44: its too simple. Its clear that selecting Cross would render crosses, and likewise for the others

src/main/java/com/lambda/mixin/world/ClientChunkManagerMixin.java
beanbag44: its a pretty niche mixin; doubt any other modules would need it. If the time comes we can always change it

PR #278, opened 2026-03-22
feat(AutoLevel): Module to automatically use xp bottles to level your player to a specific level

IceTank:
This module is useful if you use something like AutoAnvil to combine or rename items

### DIFF
--- a/src/main/kotlin/com/lambda/interaction/managers/hotbar/HotbarManager.kt
+++ b/src/main/kotlin/com/lambda/interaction/managers/hotbar/HotbarManager.kt
@@ -28,7 +28,6 @@ import com.lambda.interaction.managers.hotbar.HotbarManager.activeSlot
 import com.lambda.interaction.managers.hotbar.HotbarManager.checkResetSwap
 import com.lambda.interaction.managers.hotbar.HotbarManager.setActiveRequest
 import com.lambda.interaction.managers.hotbar.HotbarManager.setActiveSlot
-import com.lambda.interaction.material.container.containers.HotbarContainer
 import com.lambda.threading.runSafe
 import net.minecraft.item.ItemStack
 
@@ -55,9 +54,6 @@ object HotbarManager : Manager<HotbarRequest>(
 	val serverSlot get() = runSafe {
 		interaction.lastSelectedSlot
 	} ?: -1
-	val currentSlot get() = runSafe {
-		if (serverSlot != -1) HotbarContainer.slots[serverSlot] else null
-	}
 	//ToDo: something to manage stacks so the hotbar manager is strictly index based
 	private var previousStack: ItemStack? = null
 	private var swappedTicks = 0
--- a/src/main/kotlin/com/lambda/interaction/managers/hotbar/HotbarManager.kt
+++ b/src/main/kotlin/com/lambda/interaction/managers/hotbar/HotbarManager.kt
@@ -28,6 +28,7 @@ import com.lambda.interaction.managers.hotbar.HotbarManager.activeSlot
 import com.lambda.interaction.managers.hotbar.HotbarManager.checkResetSwap
 import com.lambda.interaction.managers.hotbar.HotbarManager.setActiveRequest
 import com.lambda.interaction.managers.hotbar.HotbarManager.setActiveSlot
+import com.lambda.interaction.material.container.containers.HotbarContainer
 import com.lambda.threading.runSafe
 import net.minecraft.item.ItemStack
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/lambda/interaction/managers/hotbar/HotbarManager.kt
```diff
@@ -54,6 +55,9 @@ object HotbarManager : Manager<HotbarRequest>(
 	val serverSlot get() = runSafe {
 		interaction.lastSelectedSlot
 	} ?: -1
+	val currentSlot get() = runSafe {
```
beanbag44: this is not needed. Checking the players main hand will already use the slot if changed in here.

src/main/kotlin/com/lambda/module/modules/player/AutoLevel.kt
```diff
@@ -0,0 +1,83 @@
...
+						interaction.interactItem(mc.player, Hand.MAIN_HAND)
+					}
+					timer.reset()
+				}
+			}
+		}
+	}
+
+	private fun SafeContext.withXp(block: () -> Unit) {
+		if (HotbarManager.currentSlot?.stack?.item == Items.EXPERIENCE_BOTTLE) {
```
beanbag44: just use player.mainHandStack

src/main/kotlin/com/lambda/module/modules/player/AutoLevel.kt
```diff
@@ -0,0 +1,83 @@
...
+	var targetLevel by setting("Target Level", 2, 1..100)
+	var burstAmount by setting("Burst Amount", 5, 1..100)
+	var interval by setting("Interval", 5, 0..100, unit = "ticks")
+
+	val timer = Timer()
+
+	init {
+		setDefaultAutomationConfig {
+			applyEdits {
+				hideGroup(eatConfig)
```
beanbag44: why not the other groups?

src/main/kotlin/com/lambda/module/modules/player/AutoLevel.kt
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+	}
+
+	private fun SafeContext.withXp(block: () -> Unit) {
+		if (HotbarManager.currentSlot?.stack?.item == Items.EXPERIENCE_BOTTLE) {
+			block()
+			return
+		}
+		val hotbarSlot = StackSelection.selectStack(count = 1) { isItem<ExperienceBottleItem>() }.filterSlots(HotbarContainer.slots).firstOrNull() ?: return
```
beanbag44: best to put isItem<ExperienceBottleItem>() on a new line

```kotlin
val hotbarSlot = StackSelection.selectStack(count = 1) { 
    isItem<ExperienceBottleItem>()
}.filterSlots(HotbarContainer.slots).firstOrNull() ?: return
```

src/main/kotlin/com/lambda/module/modules/player/AutoLevel.kt
```diff
@@ -0,0 +1,83 @@
...
+		HotbarRequest(
```
beanbag44: not a big deal, but i'd set the request to a variable and then check if (request.done) block() rather than using .also here

src/main/kotlin/com/lambda/module/modules/player/AutoLevel.kt
```diff
@@ -0,0 +1,83 @@
...
+import net.minecraft.util.Hand
+import kotlin.time.Duration.Companion.milliseconds
+
+
+object AutoLevel : Module(
+	name = "AutoLevel",
+	description = "Automatically uses xp bottles to level up to a certain level",
+	tag = ModuleTag.PLAYER
+) {
+	var targetLevel by setting("Target Level", 2, 1..100)
```
beanbag44: i think a rotate and angle setting would be a good addition

beanbag44 (changes requested):
Why not make it AutoXP and have an option for max level?

IceTank:
> Why not make it AutoXP and have an option for max level?

This would be less flexible. If you want to use autoxp as well as autolevel you would always have to switch modes in the module.

beanbag44:
> > Why not make it AutoXP and have an option for max level?
> 
> This would be less flexible. If you want to use autoxp as well as autolevel you would always have to switch modes in the module.

im just worried about module spam. I doubt many people would be switching back and forth that frequently? One is mainly used for pvp, the other for either renaming stuff on anvils or repairing gear?

IceTank:
I'll look at it again

beanbag44:
going to close your pr's for now as its been a while. Feel free to reopen them if they get updated of course.

PR #279, opened 2026-03-25
fix(Discord): Unfuck Discord rpc

IceTank:
This change lazy loads the Discord integration dependency, removing around a 10-second unavoidable delay when starting the game.
This also fixes the module starting the Discord RPC integration even if the module is disabled.

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/client/Discord.kt
+++ b/src/main/kotlin/com/lambda/module/modules/client/Discord.kt
@@ -39,82 +39,99 @@ import dev.cbyrne.kdiscordipc.data.activity.largeImage
 import dev.cbyrne.kdiscordipc.data.activity.smallImage
 import dev.cbyrne.kdiscordipc.data.activity.timestamps
 import kotlinx.coroutines.delay
+import java.util.concurrent.atomic.AtomicBoolean
 
 object Discord : Module(
-    name = "Discord",
-    description = "Discord Rich Presence configuration",
-    tag = ModuleTag.CLIENT,
-    enabledByDefault = true,
+	name = "Discord",
+	description = "Discord Rich Presence configuration",
+	tag = ModuleTag.CLIENT,
+	enabledByDefault = true,
 ) {
-    private val delay by setting("Update Delay", 5000L, 5000L..30000L, 100L, unit = "ms")
-    private val showTime by setting("Show Time", true, description = "Show how long you have been playing for.")
-    private val line1Left by setting("Line 1 Left", LineInfo.World)
-    private val line1Right by setting("Line 1 Right", LineInfo.Username)
-    private val line2Left by setting("Line 2 Left", LineInfo.Version)
-    private val line2Right by setting("Line 2 Right", LineInfo.Fps)
+	private val delay by setting("Update Delay", 5000L, 5000L..30000L, 100L, unit = "ms")
+	private val showTime by setting("Show Time", true, description = "Show how long you have been playing for.")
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/client/Discord.kt
```diff
@@ -67,36 +67,58 @@ object Discord : Module(
...
 
-    private suspend fun start() {
-        if (rpc.connected) return
-
-        rpc.connect()
+    private suspend fun start(): Boolean {
+        if (rpc == null) {
+            rpc = KDiscordIPC(Lambda.APP_ID, scope = EventFlow.lambdaScope)
+        }
+        if (rpc!!.connected) return true
```
beanbag44: we, if not never, almost never use !! as it voids kotlins null safety checks and opens space up for null pointer exceptions

src/main/kotlin/com/lambda/module/modules/client/Discord.kt
```diff
@@ -67,36 +67,58 @@ object Discord : Module(
             return@listenOnce true
         }
 
-        onEnable { runConcurrent { start(); handleLoop() } }
-        runConcurrent { start() }
+        onEnable {
```
beanbag44: this is the only place start() is called which will not run if the user has the module enabled on client startup as onEnable requires a SafeContext

src/main/kotlin/com/lambda/module/modules/client/Discord.kt
```diff
@@ -53,7 +53,7 @@ object Discord : Module(
     private val line2Left by setting("Line 2 Left", LineInfo.Version)
     private val line2Right by setting("Line 2 Right", LineInfo.Fps)
 
-    val rpc = KDiscordIPC(Lambda.APP_ID, scope = EventFlow.lambdaScope)
+    var rpc: KDiscordIPC? = null
```
beanbag44: could this be replaced with a lazy {}?

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/client/Discord.kt
IceTank: image

deividazevedo10-blip (approved):

src/main/kotlin/com/lambda/module/modules/client/Discord.kt
deividazevedo10-blip: Oi

deividazevedo10-blip (approved):

src/main/kotlin/com/lambda/module/modules/client/Discord.kt
deividazevedo10-blip: Oi

beanbag44: aight, i dont remember it working last i tested but i'll merge it anyway because its already broken. I mentioned that in the discord btw, just not here.

src/main/kotlin/com/lambda/module/modules/client/Discord.kt
deividazevedo10-blip: Salve

src/main/kotlin/com/lambda/module/modules/client/Discord.kt
deividazevedo10-blip: @deividazevedo10-blip

deividazevedo10-blip (commented):
Deividsantos

PR #280, opened 2026-04-02
fix(ChestContainer): Fix chest containers not recognizing some chest like containers

IceTank:
Lootable containers should include pretty much all containers in the game we would want to take items out of

### DIFF
--- a/src/main/kotlin/com/lambda/interaction/material/container/ContainerManager.kt
+++ b/src/main/kotlin/com/lambda/interaction/material/container/ContainerManager.kt
@@ -26,14 +26,14 @@ import com.lambda.event.listener.SafeListener.Companion.listen
 import com.lambda.interaction.material.ContainerSelection
 import com.lambda.interaction.material.StackSelection
 import com.lambda.interaction.material.StackSelection.Companion.select
-import com.lambda.interaction.material.container.containers.ChestContainer
+import com.lambda.interaction.material.container.containers.LootableContainer
 import com.lambda.interaction.material.container.containers.EnderChestContainer
 import com.lambda.util.BlockUtils.blockEntity
 import com.lambda.util.extension.containerStacks
 import com.lambda.util.reflections.getInstances
 import net.minecraft.block.entity.BlockEntity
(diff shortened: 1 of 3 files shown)


beanbag44 (commented):
Im not sure what this solves. Although it makes it possible to create a ChestContainer object for at least most external block entity containers in the game, the class is still called ChestContainer so the name would be misleading, and the ContainerManager still functions the same way, only creating instances for chests

IceTank:
I renamed ChestContainer to LootableContainer. This should now include all containers that can be looted.

src/main/kotlin/com/lambda/interaction/material/container/ContainerManager.kt
```diff
@@ -67,16 +67,14 @@ object ContainerManager : Loadable {
                     EnderChestContainer.update(handler.containerStacks)
                 }
 
-                is ChestBlockEntity -> {
-                    // ToDo: Handle double chests and single chests
-                    if (handler.type != ScreenHandlerType.GENERIC_9X6) return@listen
+                is LootableContainerBlockEntity -> {
```
beanbag44, line 70: you removed the todo for handling double chests but this does not fix the double chest issue. This is part of my worry about merging all external containers aside from ender chests into one. It removes flexibility for unique containers. Also, there will likely be cases where the type of container needs to be known. I think we should just stick to separate types.

beanbag44 (changes requested):

src/main/kotlin/com/lambda/interaction/material/container/ContainerManager.kt
IceTank: Double chests could be fixed by combining both inventories into one container location. I have done that with my stash movers before. I think if you need a more specific container type, you can always extend this one.

beanbag44: I agree that there should be a base open class for containers, but this class isnt suited towards that. It takes a blockPos in the constructor, which wouldnt apply well for double chests as they consist of two positions. The class also isnt open, and its a data class which cant be inherited from, etc

PR #282, opened 2026-04-10
Feature: StashMover module

beanbag44:
Adds a module that allows the user to select pull and put chests to pull and put items from and to. 

There is a pearl bot and mover bot. The pearl bot receives messages from the mover bot to pearl the account when its inventory is full. The mover bot does almost all the heavy lifting. The mover bot should have its bed set at the pull stash, and a pearl set in a contraption where the pearl bot can load it using a button press. There should be a dropper setup by the pearling station where the mover can retrieve a new pearl to set itself again.

### DIFF
--- a/src/main/kotlin/com/lambda/interaction/managers/PacketLimitHandler.kt
+++ b/src/main/kotlin/com/lambda/interaction/managers/PacketLimitHandler.kt
@@ -21,13 +21,17 @@ import com.lambda.config.groups.BuildConfig
 import com.lambda.context.Automated
 import com.lambda.event.events.TickEvent
 import com.lambda.event.listener.SafeListener.Companion.listen
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.ComparableTimeMark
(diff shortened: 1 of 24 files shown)


src/main/kotlin/com/lambda/interaction/managers/PacketLimitHandler.kt
```diff
@@ -21,45 +21,50 @@ import com.lambda.config.groups.BuildConfig
...
+			val currentTime = System.currentTimeMillis()
+			repeat((0 until packetCount).count()) {
+				handler.packetTimestamps.addLast(currentTime)
+			}
 		}
+	}
 
-	private data class LimitHandler(var maxPacketsThisTimeframe: Int, var packetsThisTimeframe: Int, var timeframe: Int) {
-		val tickTimer = TickTimer()
+	private class LimitHandler {
```
emyfops, line 57: Please dont use raw values as durations. `Duration` exist for a reason

emyfops (changes requested):

beanbag44:
🤓

PR #283, opened 2026-04-11
feat(Printer): Add flattenMode and baritone selection to printer module

IceTank:
Adds a similar flatten mode to the printer as Nuker has. It can be configured to apply to only breaking blocks, placing blocks, or both.
These features are especially useful when building mapart with printer and you want to break and place blocks in the same litematica placement area without having to use nuker.

### DIFF
--- a/src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
+++ b/src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
@@ -18,12 +18,15 @@
 package com.lambda.interaction.construction.blueprint
 
 import com.lambda.context.SafeContext
+import com.lambda.interaction.construction.simulation.result.BuildResult
 import com.lambda.threading.runSafe
 import com.lambda.util.extension.Structure
 import net.minecraft.util.math.Vec3i
+import java.util.function.Predicate
 
 data class TickingBlueprint(
-    val onTick: SafeContext.(Structure) -> Structure? = { it },
+    val buildResultPredicate: Predicate<BuildResult> = Predicate { true },
+    val onTick: SafeContext.(Structure) -> Structure? = { it }
 ) : Blueprint() {
     fun tick() =
         runSafe {
@@ -49,6 +52,6 @@ data class TickingBlueprint(
 
         fun tickingBlueprint(
             onTick: SafeContext.(Structure) -> Structure?,
-        ) = TickingBlueprint(onTick)
+        ) = TickingBlueprint(onTick = onTick)
     }
 }
--- a/src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
+++ b/src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
@@ -25,7 +25,7 @@ import net.minecraft.util.math.Vec3i
 import java.util.function.Predicate
 
 data class TickingBlueprint(
-    val buildResultPredicate: Predicate<BuildResult> = Predicate { true },
+    val buildResultPredicate: SafeContext.(BuildResult) -> Boolean = { true },
     val onTick: SafeContext.(Structure) -> Structure? = { it }
 ) : Blueprint() {
     fun tick() =
@@ -51,7 +51,8 @@ data class TickingBlueprint(
         }
 
         fun tickingBlueprint(
-            onTick: SafeContext.(Structure) -> Structure?,
-        ) = TickingBlueprint(onTick = onTick)
+            buildResultPredicate: SafeContext.(BuildResult) -> Boolean = { true },
+            onTick: SafeContext.(Structure) -> Structure?
+        ) = TickingBlueprint(onTick = onTick, buildResultPredicate = buildResultPredicate)
     }
 }
--- a/src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
+++ b/src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
@@ -18,14 +18,11 @@
 package com.lambda.interaction.construction.blueprint
 
 import com.lambda.context.SafeContext
-import com.lambda.interaction.construction.simulation.result.BuildResult
 import com.lambda.threading.runSafe
 import com.lambda.util.extension.Structure
 import net.minecraft.util.math.Vec3i
-import java.util.function.Predicate
 
 data class TickingBlueprint(
-    val buildResultPredicate: SafeContext.(BuildResult) -> Boolean = { true },
     val onTick: SafeContext.(Structure) -> Structure? = { it }
 ) : Blueprint() {
     fun tick() =
@@ -51,8 +48,7 @@ data class TickingBlueprint(
         }
 
         fun tickingBlueprint(
-            buildResultPredicate: SafeContext.(BuildResult) -> Boolean = { true },
             onTick: SafeContext.(Structure) -> Structure?
-        ) = TickingBlueprint(onTick = onTick, buildResultPredicate = buildResultPredicate)
+        ) = TickingBlueprint(onTick)
     }
 }
--- a/src/main/kotlin/com/lambda/module/modules/world/Nuker.kt
+++ b/src/main/kotlin/com/lambda/module/modules/world/Nuker.kt
@@ -20,8 +20,6 @@ package com.lambda.module.modules.world
 import com.lambda.config.AutomationConfig.Companion.setDefaultAutomationConfig
 import com.lambda.config.applyEdits
 import com.lambda.context.SafeContext
-import com.lambda.interaction.BaritoneManager
-import com.lambda.interaction.construction.blueprint.TickingBlueprint
 import com.lambda.interaction.construction.blueprint.TickingBlueprint.Companion.tickingBlueprint
 import com.lambda.interaction.construction.verify.TargetState
 import com.lambda.module.Module
@@ -30,12 +28,11 @@ import com.lambda.task.RootTask.run
 import com.lambda.task.Task
 import com.lambda.task.tasks.BuildTask.Companion.build
 import com.lambda.util.BlockUtils.blockPos
-import com.lambda.util.BlockUtils.blockState
-import com.lambda.util.BlockUtils.isNotEmpty
-import com.lambda.util.math.MathUtils.ceilToInt
 import net.minecraft.block.Blocks
 import net.minecraft.util.math.BlockPos
-import net.minecraft.util.math.Direction
+import com.lambda.util.PlayerBuildLayerUtils.FlattenMode
+import com.lambda.util.PlayerBuildLayerUtils.isInFlatten
+import com.lambda.util.PlayerBuildLayerUtils.isInBaritoneSelection
 
 object Nuker : Module(
 	name = "Nuker",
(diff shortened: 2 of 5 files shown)


src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -57,6 +70,12 @@ object Printer : Module(
 					.asSequence()
 					.filter { DataManager.getRenderLayerRange().isPositionWithinRange(it) && inSchematic(it) }
 					.associateWith { TargetState.State(schematicWorld.getBlockState(it)) }
+					.filter {
+						if (flattenModeApply == FlattenModeApply.PlaceOnly && it.value.blockState.isAir) return@filter true
+						if (flattenModeApply == FlattenModeApply.BreakOnly && !it.value.blockState.isAir) return@filter true
+						return@filter isInFlatten(it.key)
+					}
+					.filterNot { isInBaritoneSelection(it.key) == inverseSelection }
```
beanbag44: i think this is generally more confusing than isInBaritoneSelection(it.key) == !inverseSelection with a regular filter

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -75,4 +94,74 @@ object Printer : Module(
 		Class.forName("fi.dy.masa.litematica.Litematica")
 		true
 	}.getOrDefault(false)
+
+	private fun SafeContext.isInFlatten(pos: BlockPos): Boolean {
```
beanbag44: this is a direct paste of the isInFlatten function in nuker, just use the function from nuker and make it public

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -75,4 +94,74 @@ object Printer : Module(
...
+
+		val zeroedPos = pos.add(-playerPos.x, -flattenY, -playerPos.z)
+
+		return (zeroedPos.x < 0 && smartFlattenDir == Direction.EAST)
+				|| (zeroedPos.z < 0 && smartFlattenDir == Direction.SOUTH)
+				|| (zeroedPos.x > 0 && smartFlattenDir == Direction.WEST)
+				|| (zeroedPos.z > 0 && smartFlattenDir == Direction.NORTH)
+	}
+
+	private fun isInBaritoneSelection(pos: BlockPos) =
```
beanbag44: again

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -75,4 +94,74 @@ object Printer : Module(
...
+		if (!baritoneSelection) true
+		else BaritoneManager.primary?.selectionManager?.selections?.any {
+			val min = it.min()
+			val max = it.max()
+			pos.x >= min.x && pos.x <= max.x
+					&& pos.y >= min.y && pos.y <= max.y
+					&& pos.z >= min.z && pos.z <= max.z
+		} ?: false
+
+	private enum class FlattenMode {
```
beanbag44: again

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -57,6 +70,12 @@ object Printer : Module(
 					.asSequence()
 					.filter { DataManager.getRenderLayerRange().isPositionWithinRange(it) && inSchematic(it) }
 					.associateWith { TargetState.State(schematicWorld.getBlockState(it)) }
+					.filter {
+						if (flattenModeApply == FlattenModeApply.PlaceOnly && it.value.blockState.isAir) return@filter true
```
beanbag44: isAir is not a direct indication that a block there should be broken. isEmpty was created for this, as it includes breaking waterlogged blocks too.

beanbag44: even checking isEmpty is not a good solution to this. If you simulate a block and there is another block there which isnt replaceable, it will have to break the other to place the new one. A true solution would be filtering the build results in the build task so that you can guarantee breaks, placements, and interactions are accurate. If you want to do this, i would add a parameter to the build task constructors for a build result predicate which defaults to { true } which can be used to filter them before processing

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
IceTank: Having both `==` and `!` duplicates the boolean operation. Using multiple operandi instead of the intended operant is bad code design. Also if imverseSelection is always used in the context of an inverted value it should be refactored so it is not inverted by default.

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
IceTank: Duplicating these specific methods enforce a modular and isolated module design with clear separation of concerns. Abstraction specific methods to global scopes ads interaction complexities and should in my opinion be avoided if not absolutely necessary like in this case where a reuse is unlikely. Also makes it easier to copy past stuff.

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
IceTank: I'll look at that

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
IceTank: https://github.com/user-attachments/assets/c77000a5-4a1c-4b04-a5f9-8892f2df8836

src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
```diff
@@ -49,6 +52,6 @@ data class TickingBlueprint(
 
         fun tickingBlueprint(
             onTick: SafeContext.(Structure) -> Structure?,
```
beanbag44: missing param here

src/main/kotlin/com/lambda/interaction/construction/blueprint/TickingBlueprint.kt
```diff
@@ -18,12 +18,15 @@
...
 import com.lambda.context.SafeContext
+import com.lambda.interaction.construction.simulation.result.BuildResult
 import com.lambda.threading.runSafe
 import com.lambda.util.extension.Structure
 import net.minecraft.util.math.Vec3i
+import java.util.function.Predicate
 
 data class TickingBlueprint(
-    val onTick: SafeContext.(Structure) -> Structure? = { it },
+    val buildResultPredicate: Predicate<BuildResult> = Predicate { true },
```
beanbag44: Predicate is a java interface. Kotlin uses () -> <return type>. Use SafeContext.(BuildResult) -> Boolean

src/main/kotlin/com/lambda/task/tasks/BuildTask.kt
```diff
@@ -60,13 +60,15 @@ import com.lambda.util.player.SlotUtils.hotbarAndInventoryStacks
...
+import java.util.function.Predicate
 
 class BuildTask private constructor(
     private val blueprint: Blueprint,
     private val finishOnDone: Boolean,
     private val collectDrops: Boolean,
     private val lifeMaintenance: Boolean,
-    automated: Automated
+    automated: Automated,
+    private val buildResultPredicate: Predicate<BuildResult> = Predicate { true }
```
beanbag44: SafeContext.(BuildResult) -> Boolean

src/main/kotlin/com/lambda/task/tasks/BuildTask.kt
```diff
@@ -130,6 +132,7 @@ class BuildTask private constructor(
             .toList()
 
         val viableResults = results
+            .filter { buildResultPredicate.test(it) }
```
beanbag44: might as well put this after the following filter as we dont know how expensive it could be.

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -75,4 +123,74 @@ object Printer : Module(
 		Class.forName("fi.dy.masa.litematica.Litematica")
 		true
 	}.getOrDefault(false)
+
+	private fun SafeContext.isInFlatten(pos: BlockPos): Boolean {
```
beanbag44: bruh this is still all just copy and pasted from nuker. Theres no need for duplicate code

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -50,20 +66,52 @@ object Printer : Module(
 				disable()
 				return@onEnable
 			}
-			buildTask = TickingBlueprint {
+			buildTask = TickingBlueprint(onTick = {
```
beanbag44: why have onTick before the build result predicate? Also the final lambda in a function, assuming its the last param, can be placed outside the parenthesis

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -50,20 +66,52 @@ object Printer : Module(
...
+			if (inverseSelection && inSelection) return false
+			if (!inverseSelection && !inSelection) return false
+		}
+
+		if (flattenModeApply == FlattenModeApply.Both) {
+			return isInFlatten(buildResult.pos)
+		}
+		return when (buildResult) {
+			is InteractResult.Interact -> {
+				if (flattenModeApply != FlattenModeApply.PlaceOnly) {
```
beanbag44: can remove the brackets and extra return, then just put true on the same line as the if statement and add else before the isInFlatten

```kotlin
if (flattenModeApply != FlattenModeApply.PlaceOnly) true
else isInFlatten(buildResult.pos)
```

src/main/kotlin/com/lambda/module/modules/world/Printer.kt
```diff
@@ -50,20 +66,52 @@ object Printer : Module(
...
+					return true
+				}
+				isInFlatten(buildResult.pos)
+			}
+			is BreakResult.Break -> {
+				if (flattenModeApply != FlattenModeApply.BreakOnly) {
```
beanbag44: likewise

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/world/Nuker.kt
```diff
@@ -72,9 +69,9 @@ object Nuker : Module(
 					.asSequence()
 					.map { it.blockPos }
 					.filter { !world.isAir(it) }
-					.filter { flattenMode == FlattenMode.None || isInFlatten(it) }
+					.filter { flattenMode == FlattenMode.None || isInFlatten(it, flattenMode, sneakLowersFlatten, baritoneSelection, false) }
```
beanbag44: baritoneSelectionInverted is just set to false here

beanbag44 (approved):
too many changes so i pushed the fixes. Double check everything before its merged. Btw im not sure if you tested it but i dont think it worked in printer. You put the predicate in the ticking blueprint instead of the build function call, but it was never used inside the class. Please test your pr's.

IceTank:
I had the versions before tested but didn't have the chance to test the later versions. Kinda thought the changes weren't that big.

PR #288, opened 2026-05-05
fix: disable freecam on dimension change

VadimTotok:
prevents a freeze caused by conflicts between Freecam logic and world reinitialization during teleportation

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/render/Freecam.kt
+++ b/src/main/kotlin/com/lambda/module/modules/render/Freecam.kt
@@ -233,7 +233,7 @@ object Freecam : Module(
 
 		listen<PacketEvent.Receive.Pre> { event ->
 			val packet = event.packet
-			if (packet is PlayerRespawnS2CPacket) toggle()
+			if (packet is PlayerRespawnS2CPacket) disable()
 		}
 	}
 
--- a/src/main/kotlin/com/lambda/module/modules/render/Freecam.kt
+++ b/src/main/kotlin/com/lambda/module/modules/render/Freecam.kt
@@ -27,6 +27,7 @@ import com.lambda.event.events.PlayerEvent
 import com.lambda.event.events.RenderEvent
 import com.lambda.event.events.TickEvent
 import com.lambda.event.listener.SafeListener.Companion.listen
+import com.lambda.event.listener.UnsafeListener.Companion.listenUnsafe
 import com.lambda.interaction.managers.rotating.IRotationRequest.Companion.rotationRequest
 import com.lambda.interaction.managers.rotating.Rotation
 import com.lambda.interaction.managers.rotating.RotationMode
@@ -231,9 +232,8 @@ object Freecam : Module(
 			mc.crosshairTarget?.let { if (it.type != HitResult.Type.MISS) event.cancel() }
 		}
 
-		listen<PacketEvent.Receive.Pre> { event ->
-			val packet = event.packet
-			if (packet is PlayerRespawnS2CPacket) disable()
+		listenUnsafe<PacketEvent.Receive.Pre> { event ->
+			if (event.packet is PlayerRespawnS2CPacket) disable()
 		}
 	}
 
--- a/src/main/kotlin/com/lambda/module/modules/render/Freecam.kt
+++ b/src/main/kotlin/com/lambda/module/modules/render/Freecam.kt
@@ -22,6 +22,7 @@ import com.lambda.config.AutomationConfig.Companion.setDefaultAutomationConfig
 import com.lambda.config.applyEdits
 import com.lambda.context.SafeContext
 import com.lambda.event.events.MovementEvent
+import com.lambda.event.events.PacketEvent
 import com.lambda.event.events.PlayerEvent
 import com.lambda.event.events.RenderEvent
 import com.lambda.event.events.TickEvent
@@ -55,6 +56,7 @@ import com.lambda.util.world.raycast.RayCastUtils.orMiss
 import net.minecraft.client.network.ClientPlayerEntity
 import net.minecraft.client.option.Perspective
 import net.minecraft.entity.player.PlayerEntity
+import net.minecraft.network.packet.s2c.play.PlayerRespawnS2CPacket
 import net.minecraft.util.hit.BlockHitResult
 import net.minecraft.util.hit.HitResult
 import net.minecraft.util.math.BlockPos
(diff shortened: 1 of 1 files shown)


IceTank:
There is a different pr that fixes the same issue without disabling the freecam module. That one should be preferred.

VadimTotok:
> There is a different pr that fixes the same issue without disabling the freecam module. That one should be preferred. 

All popular clients use exactly this type of protection against game crashes.

Disabling the freecam module is a lightweight and safe fix - it prevents the crash. I've reviewed the solution from another pr, and here's why it's worse: the game doesn't crash, but you lose a lot of performance. Loading the world takes twice as long or even more - it's very slow and inefficient.

My solution doesn't lose performance and does exactly the same thing every popular client does to solve this problem.

IceTank:
> > There is a different pr that fixes the same issue without disabling the freecam module. That one should be preferred.
> 
> All popular clients use exactly this type of protection against game crashes.
> 
> Disabling the freecam module is a lightweight and safe fix - it prevents the crash. I've reviewed the solution from another pr, and here's why it's worse: the game doesn't crash, but you lose a lot of performance. Loading the world takes twice as long or even more - it's very slow and inefficient.
> 
> My solution doesn't lose performance and does exactly the same thing every popular client does to solve this problem.

Is this response AI-generated? I didn't notice any performance impact. 

Disabling modules without user input is very annoying if you have multiple modules bound to the same key. We aim not to disable any modules without explicit user input.

VadimTotok:
> > > There is a different pr that fixes the same issue without disabling the freecam module. That one should be preferred.
> > 
> > All popular clients use exactly this type of protection against game crashes.
> > 
> > Disabling the freecam module is a lightweight and safe fix - it prevents the crash. I've reviewed the solution from another pr, and here's why it's worse: the game doesn't crash, but you lose a lot of performance. Loading the world takes twice as long or even more - it's very slow and inefficient.
> > 
> > My solution doesn't lose performance and does exactly the same thing every popular client does to solve this problem.
> 
> Is this response AI-generated? I didn't notice any performance impact. 
> 
> Disabling modules without user input is very annoying if you have multiple modules bound to the same key. We aim not to disable any modules without explicit user input.

which module will the user want to bind to the same button that freecam is binding to? in my case, the loading of the world was very long, the game did not crash, but the loading time was very long. I checked it several times. I have a good computer. if you disable the freecam module, everything will work as it should in good clients. because it makes no sense to leave the freecam module enabled! This is nonsense!

IceTank:
Ah is see what you mean now. The long loading screen was a bug in the implementation. But that should be fixed now.

Freecam has a feature to stay in a position relative to the player. Some people might toggle freecam and something like autowalk or highway tools with the same key.

VadimTotok:
> Ah is see what you mean now. The long loading screen was a bug in the implementation. But that should be fixed now.
> 
> Freecam has a feature to stay in a position relative to the player. Some people might toggle freecam and something like autowalk or highway tools with the same key. 

okay, but it doesn't fit the dimension change scenario. therefore, turning off freecam is still the best decision that can be made, and it will not have consequences. I'm not suggesting that the player randomly disable the freecam module. I highly doubt that players need to leave the freecam on when they have changed the dimension.

IceTank:
I don't understand what you mean by dimension change

VadimTotok:
> I don't understand what you mean by dimension change

ok.

by dimension, I mean nether, overworld, the end. This pr is aimed at correcting the problem that if a player passes through a portal from the world to the nether or vice versa and out of curiosity decides to fly using the freecam module and does not have time to turn it off during the transition from one dimension to another, the game crashes. My pr was created in order to fix this situation so that the game does not crash by simply disabling the freecam module! Why the hell would he need it after moving from one dimension to another?!

IceTank:
Switching dimensions doesn't cause any crashing for me. Did you test my branch?

VadimTotok:
if everything is working fine in ur implementation, feel free to close this pr :^)

PoTheMagicDragon:
@IceTank Has that PR you mentioned been merged/ do you feel good about closing this PR?

beanbag44:
I think this seems to be the best solution. The amount of code required to ensure Freecam can stay enabled through a dimension change seems too like way much for the job, especially as i think most people would expect freecam to disable so they can walk out of the portal and continue playing normally without having to disable it manually. I guess maybe people would want it to persist so they can compare relative positions or something but the other pr is still in draft too.

@VadimTotok the only things i'd change about the pr are making the listener an unsafe listener so it works regardless if theres a player or not. A safe listener might not trigger if the player dies or something. I'd also change the toggle() call to a disable() call and probably just inline the packet value.

Sorry its taken me 2 months to comment on this. I really gotta get better at checking and reviewing pr's

beanbag44 (changes requested):
the listener is still a safe listener. This might not trigger if the player dies or gets removed from the world before the packet is received

VadimTotok:
why did you add my code, but not add me to the list of contributors and not close this pr? I created this pr ALMOST THREE MONTHS AGO. You're not doing a very good job

beanbag44:
> why did you add my code, but not add me to the list of contributors and not close this pr? I created this pr ALMOST THREE MONTHS AGO. You're not doing a very good job

I don't think i did add your code. If I'm mistaken, can you link the commit where it was added? I am sorry it's taken so long to get merged, i wasn't sure if there was going to be a conclusion about yours and icetanks fixes but i should have seen it was over sooner. I'm ready to merge this pr, the listener just needs to be unsafe so it definitely triggers correctly

VadimTotok:
блять look at the changes you've made and the changes I've suggested.

https://github.com/lambda-client/lambda/pull/288/files/4fe162a4746e997ac13e8604faec71d4b659ddb0
https://github.com/lambda-client/lambda/pull/288/commits/4fe162a4746e997ac13e8604faec71d4b659ddb0
https://github.com/lambda-client/lambda/pull/288/commits/1bcc0f4e071dffa25732e1ff3e4e785eb86547f3

You asked me to replace change the toggle() call to a disable() call and then you just made changes to the code instead of merging.

почему вы такие сложные..

beanbag44:
im confused. I didn't change any of your code or merge it into the main branch. Both commits on this pr are from your account and the pr is still open. If you look at Freecam.kt on the main branch, none of this code has been added yet. Am i missing something?

beanbag44:
@VadimTotok I'm not sure what the issue is, but i'm going to make the fix myself as i have write access so it can be merged.

PR #291, opened 2026-05-11
Update gradle wrapper and loom

IceTank:
because


emyfops:
I would prefer updating when changing version. Unless there is a bug with a specific version there is no need to change it

IceTank:
I wanted to add a Xaeros Plus integration, but the current version of Gradle is incompatible. Without this, lambda won't be able to integrate with Xaeros.

PoTheMagicDragon:
@IceTank Have you been running with an updated gradle/loom since this PR? And if so, has it been stable?

PR #292, opened 2026-05-12
fix: ContainerTransferTask crash when transferring 0-count selection

### LINKED ISSUE #226: Bug: Pickup pickaxe error
### What happened?

highwaytools try to pickup a pickaxe from shulker box, but open a shulker box without a pickaxe and throw error

[log.txt](https://github.com/user-attachments/files/24789142/log.txt)

### What should have happened?

Pickup pickaxe form shulker box, add drop empty shulker box or auto fallback to other shulker box?

### Version

1.21.11 Build54

### Log Output

_No response_

we-02:
Fixes #226

### DIFF
--- a/src/main/kotlin/com/lambda/interaction/construction/simulation/result/results/GenericResult.kt
+++ b/src/main/kotlin/com/lambda/interaction/construction/simulation/result/results/GenericResult.kt
@@ -99,6 +99,7 @@ sealed class GenericResult : BuildResult() {
 
         context(task: Task<*>, _: AutomatedSafeContext)
         override fun resolve() {
+            if (neededSelection.count == 0) return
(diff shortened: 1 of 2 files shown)


beanbag44:
Sorry its taken me a while to get around to this. The ContainerTransferTask as a whole is a big mess lol, i remember i was very sleep deprived when i wrote it so its probably going to be rewritten soon with some new changes. Im not sure about the success() call when checkFail is called as its not actually succeeded. But yeah that function seems to be broken from the beginning. I believe i wrote it initially to only fail if failIfNoMaterial was true and simply return otherwise and wait for the material.

For the stack size 0 check, the current system actually allows for stack size 0 selections for times when you want to select empty slots, and in the case of #226 the error was more of a design flaw which i believe is fixed now with the soft failing functionality in tasks.

PR #294, opened 2026-05-15
Enable ModuleNotifier by default

IceTank:
Sets ModuleNotifier to enabled by default


IceTank:
Update to use buildText

PR #295, opened 2026-05-16
Extend InventoryTweaks functionallity to containers

IceTank:
Allow InventoryTweaks to take shulkers from containers and automatically place them back


beanbag44:
There are plans to fix the more generalized recursive task system in place for this functionality

PR #298, opened 2026-06-02
Improvement: Config System

beanbag44:
Replaces the old group system with annotation-based grouping and adds a second type. The old group name was used for tabs, now, @ Group(*groups*) is used for actual drop-down groups, and @ Tab(*tabs*) is used for tabs (the old group functionality). With the more strongly integrated layering, groups and tabs are now serialized, allowing duplicate setting names within the same config as long as they are separated by a layer.

The old system had a patchy style of grouping settings; however, now it is supported directly inside the Config's, with systems in place to ensure they're easy to create and use. Previously, you had to include a prefix parameter in every SettingGroup and apply it to the start of each settings name to ensure if you had two groups of the same kind, the names wouldnt conflict. You also had to use a helper .index() call at the end of every setting so that when using the ConfigEditor api, it could distinguish the settings inside and out of any given group. Now, SettingGroup has been renamed to ConfigBlock to avoid confusion with the grouping system, and none of these problems are present.

Lots of the issues previously encountered were fixed mainly by the annotations enforcing use of reflections for gathering setting metadata. This is what helped make ConfigBlock integration much easier; however, it tightened the restrictions as to where you could register settings from. All settings must be class-level values or variables to be found via reflections, and subsequently registered.

Removes all SettingCore child classes and moves the logic into Setting child classes instead. Im not sure why i initially put the logic in the SettingCore classes as it made handling listeners and the automation configs a lot more difficult, and in some cases outright causing unfixable problems.

Adds the ability to save "Property"'s just like settings. Property's have all the basic functions of settings. A setting core, an isModified getter, a name, etc. They do not have a gui builder, or a command, or any of the extra functionality of settings. They are also generically typed. This means that you can create a property of any type, and have it save and load, to and from the config.

Configs now save two sections. Settings, and Properties. They initially had the Properties object on the same level as the regular settings, however, i separated them to avoid conflicting namespace.

Migrates from Gson to Jackson as it has support for updating already initialized objects rather than creating new instances. This allows us to keep our singleton-based configs while moving away from serialization and deserialization within the classes themselves. This means the Jsonable interface is also removed. Alongside the Jackson migration, we also have a custom Serializers and Deserializers class that searches the objects supertypes for the closest relevant (de)serializer, e.g., so we dont have to write them for each module. These serializers extend FallbackSerializer or FallbackDeserializer. The standard serializers still extend the regular Serializer and Deserializer classes.

Configs now only save modified settings. Settings have an open isModified value so settings like collection settings can check size and match elements to see if anythings changed


src/main/java/com/lambda/mixin/render/WorldRendererMixin.java
```diff
@@ -93,16 +93,15 @@ private boolean renderSetupTerrainModifyArg(boolean spectator) {
         return Freecam.INSTANCE.isEnabled() || CameraTweaks.INSTANCE.isEnabled() || spectator;
     }
 
-    @Inject(method = "fillEntityRenderStates", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/entity/EntityRenderManager;shouldRender(Lnet/minecraft/entity/Entity;Lnet/minecraft/client/render/Frustum;DDD)Z", shift = At.Shift.BEFORE), locals = LocalCapture.CAPTURE_FAILHARD)
-    private void injectFillEntityRenderStates(Camera camera, Frustum frustum, RenderTickCounter tickCounter,
-            WorldRenderState renderStates, CallbackInfo ci, Vec3d vec3d, double d, double e, double f,
-            TickManager tickManager, boolean bl, Iterator var14, Entity entity) {
+    @SuppressWarnings("DiscouragedShift")
+    @Inject(method = "fillEntityRenderStates", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/entity/EntityRenderManager;shouldRender(Lnet/minecraft/entity/Entity;Lnet/minecraft/client/render/Frustum;DDD)Z", shift = At.Shift.BEFORE))
```
IceTank, line 97: Do you need this before the shift? I thought mixins were inserted before the target operation anyway?

PoTheMagicDragon:
Manager -> Handler Same thing. Might not be worth renaming so that the PR is easier to process. 

Edit after looking through some more: Ya... this would be way easier to process if all the renaming happened in a different PR so that I can find the edits relevant to the functionality changes

src/main/java/com/lambda/mixin/render/WorldRendererMixin.java
beanbag44: idk, i didnt wanna change it in case it breaks stuff and its a small thing so i just added the warning suppression

beanbag44:
> Manager -> Handler Same thing. Might not be worth renaming so that the PR is easier to process.
> 
> Edit after looking through some more: Ya... this would be way easier to process if all the renaming happened in a different PR so that I can find the edits relevant to the functionality changes

yah i can probably move the renamed stuff to a different pr. Didnt really think about that when i changed them. As for the reason for changing them, the managers are things like BreakManager, InteractManager, RotationManager, etc. Its mostly as they take and manage requests. The handlers just handle things and provide util style functionality

PoTheMagicDragon:
Ah, makes sense

src/main/kotlin/com/lambda/command/CommandHandler.kt
```diff
@@ -40,7 +40,7 @@ import kotlin.math.min
 
 
 object CommandHandler {
-    private const val ERROR_PADDING = 10
+    private const val ErrorPadding = 10
```
PoTheMagicDragon: Is it not convention in kotlin to have const values be shout case?

src/main/kotlin/com/lambda/config/ConfigEditor.kt
```diff
@@ -15,131 +15,264 @@
...
-import com.lambda.util.NamedEnum
+import com.lambda.config.entries.Property
+import com.lambda.config.entries.Setting
+import com.lambda.context.SafeContext
 import kotlin.reflect.KProperty0
 import kotlin.reflect.jvm.isAccessible
 
 @DslMarker
-annotation class SettingEditorDsl
+annotation class ConfigEditorD5l
```
PoTheMagicDragon, line 29: Why D5l?

PoTheMagicDragon (commented):
I'm not sure I'm qualified to give feedback on this pr tbh. I don't understand enough of the code base to fully understand what is happening, so from what I've seen, the changes seem good, and I'm just going to trust that you are confident in the changes.

src/main/kotlin/com/lambda/config/ConfigEditor.kt
beanbag44: oh yeah i should have specified in the pr. The dsl markers in most of the codebase dont actually do anything compilation wise. We just use them because they colour code the api's lol

beanbag44: the colour of the dsl marker is based on the hash of the name, so we sometimes replace s with 5 to change the hash to get a better colour

beanbag44:
> I'm not sure I'm qualified to give feedback on this pr tbh. I don't understand enough of the code base to fully understand what is happening, so from what I've seen, the changes seem good, and I'm just going to trust that you are confident in the changes.

its all good, i'll merge it shortly. Thanks for taking a look anyway tho

src/main/kotlin/com/lambda/command/CommandHandler.kt
beanbag44: it is, but when changing to the annotation based grouping, PascalCase looked so much better and was easier to type, so i figured it was worth switching. Its something thats easy to revert in the future if it turns out to be a bad decision

PR #302, opened 2026-06-14
Feat: Open ClickGui on More Screens

PoTheMagicDragon:
### Description

Makes it so that the click gui can be opened on more screens. That way the player can modify settings when they aren't logged in. It's particularly relevant to the auto disconnect flow

Note: There are several comments in this PR because there are several places where it isn't super obvious what the code is doing. I'm not sure if it needs restructuring or if the comments are the best approach and I'm curious if you have any ideas.


src/main/kotlin/com/lambda/module/modules/combat/autodisconnect/AutoDisconnectScreen.kt
```diff
@@ -33,7 +34,7 @@ import net.minecraft.text.Text
 import kotlin.math.min
 
 class AutoDisconnectScreen(private val details: DisconnectDetails) :
-    Screen(Text.literal("Disconnected: ").append(details.reason)) {
+    Screen(Text.literal("Disconnected: ").append(details.reason)), OverlayBackgroundScreen {
```
beanbag44: for classes like this that take a parameter, we usually format it like this:

```kotlin
class AutoDisconnectScreen(
    private val details: DisconnectDetails
) : Screen(Text.literal("Disconnected: ").append(details.reason)), OverlayBackgroundScreen {
}
```

for classes that have a lot of interfaces, we just move the rest of the interfaces onto a new line after the comma:

```kotlin
class AutoDisconnectScreen(
	private val details: DisconnectDetails
) : Screen(Text.literal("Disconnected: ").append(details.reason)),
    OverlayBackgroundScreen {
}
```

src/main/kotlin/com/lambda/gui/components/ClickGuiLayout.kt
```diff
@@ -71,10 +76,42 @@ object ClickGuiLayout : Loadable, Config(
...
+	init {
+		// The keybind's own onPress runs through a SafeListener, which only fires in-game.
+		// Register unsafe listeners so the GUI can also be toggled from menu screens
+		// (title, multiplayer, world-select) where there is no SafeContext.
+		listenUnsafe<ButtonEvent.Keyboard.Press> { event -> handleMenuToggle(event) }
+		listenUnsafe<ButtonEvent.Mouse.Click> { event -> handleMenuToggle(event) }
+	}
+
+	private fun handleMenuToggle(event: ButtonEvent) {
+		// In-game is handled by the keybind's SafeListener onPress; avoid double-toggling.
```
beanbag44: its probably just worth removing the safe context from bind listeners and other setting callbacks like onSelect and onDeselect in CollectionSetting. These were only provided with safe context because they could only really be changed while in game. Now that thats not the case, we should leave it up to the dev to decide if they need to wrap their code in runSafe {}

src/main/kotlin/com/lambda/gui/LambdaScreen.kt
```diff
@@ -23,10 +23,46 @@ import net.minecraft.client.gui.screen.Screen
...
 
+    override fun renderBackground(context: DrawContext, mouseX: Int, mouseY: Int, delta: Float) {
+        if (parentScreen == null) {
+            // In-game: keep vanilla behavior (blur + darken the rendered world).
+            super.renderBackground(context, mouseX, mouseY, delta)
+            return
+        }
+        // Draw a panorama backdrop so menus that rely on renderBackground for their
+        // background (e.g. MultiplayerScreen) aren't black, then the parent's widgets.
+        renderPanoramaBackground(context, delta)
```
beanbag44: i noticed the panorama seemed to spin faster when the click gui was toggled

beanbag44 (changes requested):
This is really nice to use in game. People have been asking for this feature for a while so its very much appreciated. Just a few things

src/main/kotlin/com/lambda/gui/components/ClickGuiLayout.kt
PoTheMagicDragon: Does this mean we need to go back through all the modules and figure out if any of their code needs to be adjusted in regards to safe vs unsafe keybinds?

beanbag44: Unfortunately yes. I get the pain, the config rework required way too much of this lol. I can do it if you dont wanna go through the hassle though

PoTheMagicDragon: It's more that I worry I don't fully understand how everything is intertwined. Give me a bit to wrap my head around how the keybind system works before I decide. I'd rather learn and take it on than have you do it, but I also don't want to make a giant pile of trash.

I'd mostly just be looking at keybind logic, and .onPress/ onSelect/ onDeselect callbacks, right?

beanbag44: yeah every callback for every setting. Theres not as many as you'd think. Theyre also all in companion objects in the setting classes. The way i'd do it, is just find every use case of the callbacks with intellij's find usages or middle-clicking/ctrl-clicking and make sure they work without safe context, or update them to wrap the required code in runSafe. Then just change the function signature to remove SafeContext

src/main/kotlin/com/lambda/gui/LambdaScreen.kt
PoTheMagicDragon: Fixed

PR #305, opened 2026-06-24
Feature: AutoElytraSwap

beanbag44:
Automatically swaps the users elytra when attempting to glide. This isn't just for standard flying either, it swaps to chestplate for fake fly, and has a couple extra settings. ElytraFly Only, so it won't swap your chestplate/elytra when regularly jumping and attempting to glide, and a glide delay to only start gliding after a certain amount of time post-swapping.


src/main/kotlin/com/lambda/module/modules/combat/AutoArmor.kt
```diff
@@ -57,12 +58,16 @@ object AutoArmor : Module(
 	private val feetProtection by setting("Preferred Feet Protection", Protection.Protection)
 	private val ignoreBinding by setting("Ignore Binding", true, "Ignores curse of binding armor pieces")
 
-	val sorter = compareByDescending<Slot> {
+	private var tickedThisTick = false
+
+	var overriddenElytraPriority: Boolean? = null
+
+	val SORTER = compareByDescending<Slot> {
```
PoTheMagicDragon, line 65: Normally I'd only make constants all upper case

src/main/kotlin/com/lambda/module/modules/combat/AutoArmor.kt
```diff
@@ -102,41 +107,44 @@ object AutoArmor : Module(
...
+			tickedThisTick = false
+		}
+	}
 
-			val swaps = mutableListOf<Pair<Slot, Slot>>()
-			armorSlots.forEach { equipped ->
-				val new = swappable.find { new ->
-					equipped.canInsert(new.stack) && sorter.compare(equipped, new) > 0
-				} ?: return@forEach
+	fun SafeContext.tick() {
```
PoTheMagicDragon, line 118: Should there be an elytra manager instead of having auto armor, auto elytra swap, & elytra fly all tangled together?

src/main/kotlin/com/lambda/module/modules/player/AutoElytraSwap.kt
```diff
@@ -0,0 +1,228 @@
...
+
+object AutoElytraSwap : Module(
+	name = "AutoElytraSwap",
+	description = "Automatically equips an elytra when attempting to glide, and removes it afterward",
+	tag = ModuleTag.PLAYER
+) {
+	@JvmStatic val elytraFlyOnly by setting("ElytraFly Only", false, "Only swaps the chest piece when the ElytraFly module is enabled and gliding")
+	private val glideDelay by setting("Glide Delay", 0, 0..20, 1, "The delay, in ticks, between swapping to elytra, and starting to glide", unit = " ticks")
+
+	val ELYTRA_SELECTION =
```
PoTheMagicDragon: Same thing about constants and case

src/main/kotlin/com/lambda/module/modules/player/AutoElytraSwap.kt
```diff
@@ -0,0 +1,228 @@
...
+		selectStack {
+			sortByDescending {
+				it.getEnchantment(Enchantments.UNBREAKING)
+			}.thenByDescending {
+				it.getEnchantment(Enchantments.MENDING)
+			}
+			isItem(Items.ELYTRA)
+				.and { it.damage < it.maxDamage }
+		}
+	val CHESTPLATE_SELECTION =
```
PoTheMagicDragon: Same thing about constants and case

src/main/kotlin/com/lambda/util/player/PlayerUtils.kt
```diff
@@ -19,57 +19,74 @@ package com.lambda.util.player
...
+import net.minecraft.entity.effect.StatusEffects
 import net.minecraft.entity.projectile.FireworkRocketEntity
 import net.minecraft.item.ItemStack
 import net.minecraft.network.packet.c2s.play.HandSwingC2SPacket
 import net.minecraft.util.Hand
 import net.minecraft.world.GameMode
 
-const val FAKE_PLAYER_ID = -2024-4-20
+object PlayerUtils {
+    const val FAKE_PLAYER_ID = -2024-4-20
```
PoTheMagicDragon, line 35: I like having these put into a class instead of free floating in a file

PoTheMagicDragon (commented):
I don't know enough about elytra logic to comment on that, so my comments are all naming/ code structure related

src/main/kotlin/com/lambda/module/modules/combat/AutoArmor.kt
beanbag44: i think it also applies to constant-style values, as long as theyre completely immutable. In the example you sent me, it included regex too which you cant use const on either

src/main/kotlin/com/lambda/module/modules/combat/AutoArmor.kt
beanbag44: yeah youre right. I never meant for AutoElytraSwap to end up handling so much, it kind of just happened and i didnt really think too much about moving it out. I'll make an ElytraHandler object somewhere

PoTheMagicDragon:
Looks good now

PR #309, opened 2026-07-04
add Pause When Eating to Breaking and Interact. Packet mine still progresses while your eatting but wont actually finish to your done eatting.This is because you cant eat when placing and and stuff.

chinesedog99:


chinesedog99:
Sorry about my spelling im like half asleep :sob:

src/main/kotlin/com/lambda/interaction/managers/breaking/BreakManager.kt
```diff
@@ -747,6 +748,7 @@ object BreakManager : Manager<BreakRequest>(
```
beanbag44, line 748: i'd move the eating check here. The current check cancels swinging at the end, while allowing swinging when constant so it would be more consistent

src/main/kotlin/com/lambda/interaction/managers/breaking/BreakManager.kt
```diff
@@ -815,7 +817,7 @@ object BreakManager : Manager<BreakRequest>(
 
 		val blockState = blockState(ctx.blockPos)
 		val progress = blockState.calcBreakDelta(ctx.blockPos)
-		val instantBreakable = progress >= info.getBreakThreshold()
+		val instantBreakable = progress >= info.getBreakThreshold() && !(breakConfig.pauseWhenEating && player.isEating)
```
beanbag44, line 820: i'd probably move the check from here to just before startBreaking is actually called. This would also catch rebreaks too, and we dont need to start instant breakable blocks before they should be broken as they break in one tick anyway

beanbag44 (changes requested):

beanbag44:
nah ur good lol. I think this would be better as a general build config setting as i'd imagine if someone wants to pause breaking when eating, they'd also want to pause other interactions like placing too.

PR #310, opened 2026-07-05
Custom shader loader + outline shaders

sophimoo:
Bit of a bloated pr has a few responsibilities but adds a custom fragment shader loader. 
Currently only brings support for outline/fill shader for ESP


sophimoo:
closing for now, will work on a rewrite as im not happy with this

PR #316, opened 2026-07-15
feat: create EndermanLook

leonetics:
Stares down endermen or keeps you from staring at them, depending on the mode.


src/main/kotlin/com/lambda/module/modules/player/EndermanLook.kt
```diff
@@ -0,0 +1,100 @@
...
+						?: closest(rest.filter { isStaringAt(it) })?.let { lookAway() }
+				}
+			}
+		}
+	}
+
+	private fun SafeContext.stareAt(enderman: EndermanEntity) =
+		rotationRequest { rotation(player.eyePos.rotationTo(enderman.eyePos)) }.submit()
+
+	private fun SafeContext.lookAway() =
```
beanbag44: SafeContext can be removed here

beanbag44 (changes requested):
This is really good, just one tiny thing. I also noticed that if you have Stun Hostiles enabled and try to fly with an elytra or something it will launch you straight at them lol. I think an option to disable the functionality when gliding as a whole would be a good addition

leonetics:
fin changes

PR #317, opened 2026-07-15
Feat:Adds Line Lock Module

PoTheMagicDragon:
### Description
An alternative to rotation lock. Instead of locking your rotation to a certain angle, it locks it towards a line. That way when you bump pigmen on the highway, you realign to the middle instead of getting pushed off to the side.

This comment explains the functionality pretty clearly
```
		//  When enabled we lock onto a straight "rail" and steer the camera down it:
		//
		//        lineOrigin ●─────────●──────────►  lineDirection (the way the rail points)
		//                            ╱ closestPoint (rail spot nearest you)
		//                    you ●──╯                → we aim a bit further down the rail
		//                                              (lookaheadPoint) so you curve back on
```

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
+++ b/src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
@@ -19,6 +19,7 @@ package com.lambda.module.modules.player
 
 import com.lambda.config.ConfigEditor.editSetting
 import com.lambda.config.ConfigEditor.hideAllExcept
+import com.lambda.config.Group
 import com.lambda.config.automation.AutomationConfig.Companion.setDefaultAutomationConfig
 import com.lambda.config.withEdits
 import com.lambda.context.SafeContext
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		//  Every tick: find the rail spot nearest you, pick a point a little ahead of it,
+		//  and turn to face that point.
+		listen<TickEvent.Pre> {
+			if (player.velocity == Vec3d.ZERO) return@listen
+			adjustYaw()
+		}
+
+		immediateRenderer("HighwayFollower Renderer") {
```
beanbag44: best to change this to LineLock Immediate Renderer

src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
```diff
@@ -0,0 +1,165 @@
...
+import kotlin.math.roundToInt
+import kotlin.math.sin
+
+@Suppress("unused")
+object LineLock : Module(
+	name = "LineLock",
+	description = "Locks the player's yaw to a line and steers toward a point ahead on the line",
+	tag = ModuleTag.Companion.PLAYER,
+) {
+	private const val YAW_SNAP_INCREMENT = 45.0     // snap the axis to the nearest 45° heading
```
beanbag44: i think it'd be nice to have some of these as settings. For example, bounce efly has the step as a setting so you can ensure that no matter what angle the highway is built at, it will always be configurable to snap in the right direction

src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
```diff
@@ -0,0 +1,165 @@
...
+import kotlin.math.cos
+import kotlin.math.max
+import kotlin.math.roundToInt
+import kotlin.math.sin
+
+@Suppress("unused")
+object LineLock : Module(
+	name = "LineLock",
+	description = "Locks the player's yaw to a line and steers toward a point ahead on the line",
+	tag = ModuleTag.Companion.PLAYER,
```
beanbag44: unused Companion reference

beanbag44 (changes requested):
The movement based snapping probably doesnt need to be used here. If the direction is defined by the starting rotation and snapping, then that can be used to angle everything. The only reason its done using movement in obstacle passing efly's is because the user might not be using a rotation lock. Its not needed here as the module is rotation locking

src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
PoTheMagicDragon: What do you mean?

PoTheMagicDragon:
> The movement based snapping probably doesnt need to be used here. If the direction is defined by the starting rotation and snapping, then that can be used to angle everything. The only reason its done using movement in obstacle passing efly's is because the user might not be using a rotation lock. Its not needed here as the module is rotation locking

What do you mean by movement based snapping?

src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
beanbag44: ModuleTag.Companion.PLAYER can be replaced with ModuleTag.PLAYER

beanbag44:
> What do you mean by movement based snapping?

The direction calculation based on line origin and current player position

PoTheMagicDragon:
> The direction calculation based on line origin and current player position

Still not sure I follow. What function/line number are you talking about?

beanbag44:
> Still not sure I follow. What function/line number are you talking about?

Reading it again i was mistaken, my bad

beanbag44:
noticed there was a reference to ConfigEditor.hideAllExcept though. Best to directly import hideAllExcept and remove the ConfigEditor. prefix

src/main/kotlin/com/lambda/module/modules/player/LineLock.kt
PoTheMagicDragon: Which ones? I left a lot of these out to keep the interface simple so there's a shallow learning curve for using it.

beanbag44: tbh i think all the constants that could be configurable should be settings. Perhaps the simple settings would be at base level, and then you could have an Advanced group under then base settings for more niche changes

PoTheMagicDragon: I like the way you think

PoTheMagicDragon:
I think all issues have been addressed

PR #318, opened 2026-07-16
Adds a better auto walk

PoTheMagicDragon:
Allows movement in all directions and adds options to pause while mining/ placing

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
@@ -59,7 +59,7 @@ object AutoWalk : Module(
 	@Group(PAUSING_GROUP) private val pauseWhilePlacing by setting("Pause While Placing", false, "Pauses walking while placing blocks or when places are queued - only works with blocks placed by a lambda module")
 	@Group(PAUSING_GROUP) private val pauseIfPlacedLastTick by setting("Pause If Placed Last Tick", false, "Pauses walking on the tick after placing a block - Works with all placing")
 
-	private var didPlaceLastTick = false
+	private var placedLastTick = false
 
 	private val SafeContext.breakQueued
 		get() = interaction.isBreakingBlock ||
@@ -73,15 +73,15 @@ object AutoWalk : Module(
 			InteractManager.blockedPositions.isNotEmpty()
 
 	init {
-		onDisable { didPlaceLastTick = false }
+		onDisable { placedLastTick = false }
 
 		listen<PacketEvent.Send.Pre> { event ->
-			if (event.packet is PlayerInteractBlockC2SPacket) didPlaceLastTick = true
+			if (event.packet is PlayerInteractBlockC2SPacket) placedLastTick = true
 		}
 
 		listen<MovementEvent.InputUpdate> { event ->
-			val placedLastTick = didPlaceLastTick
-			didPlaceLastTick = false
+			val placedLastTick = placedLastTick
+			AutoWalk.placedLastTick = false
 
 			if (pauseIfPlacedLastTick && placedLastTick ||
 				pauseWhileMining && breakQueued ||
--- a/src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
@@ -90,7 +90,7 @@ object AutoWalk : Module(
 
 			val input = event.input
 			if (walkForward || walkBackward) input.forward = (walkForward.toFloat() - walkBackward.toFloat())
-			if (strafeLeft || strafeRight) input.strafe = (strafeRight.toFloat() - strafeLeft.toFloat())
+			if (strafeLeft || strafeRight) input.strafe = (strafeLeft.toFloat() - strafeRight.toFloat())
 			if (sneak || sprint) input.update(sneak = sneak || input.sneaking, sprint = sprint || input.sprinting)
 
 			if (limitSpeed) input.movementVector = Vec2f(input.strafe, input.forward).normalize().multiply(speed)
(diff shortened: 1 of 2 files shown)


greggh:
This would be very helpful!

src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
```diff
@@ -17,28 +17,79 @@
...
 @Suppress("unused")
 object AutoWalk : Module(
 	name = "AutoWalk",
-	description = "Automatically makes your character walk forward",
+	description = "Automatically walks in the configured direction when certain conditions are met",
 	tag = ModuleTag.MOVEMENT,
 ) {
-	val limitSpeed by setting("Limit Speed", false)
-	val speed by setting("Speed", 0.5, 0.1..1.0, 0.05) { limitSpeed }
+	@Group("Movement") private val walkForward by setting("Forward", true, "Automatically walks forward")
```
beanbag44: for groups and tabs we always store them in a constant. Probably best to put them at the top of the object too so its easy to see what categories of settings there are in the module

src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
```diff
@@ -17,28 +17,79 @@
...
+	@Group("Movement") private val sprint by setting("Sprint", false, "Automatically sprints")
+
+	@Group("Speed") private val limitSpeed by setting("Limit Speed", false)
+	@Group("Speed") private val speed by setting("Speed", 0.5, 0.1..1.0, 0.05) { limitSpeed }
+
+	@Group("Pausing") private val pauseWhileMining by setting("Pause While Mining", false, "Pauses walking while breaking blocks or when breaks are queued")
+	@Group("Pausing") private val pauseWhilePlacing by setting("Pause While Placing", false, "Pauses walking while placing blocks or when places are queued - only works with blocks placed by a lambda module")
+	@Group("Pausing") private val pauseIfPlacedLastTick by setting("Pause If Placed Last Tick", false, "Pauses walking on the tick after placing a block - Works with all placing")
+
+	private var didPlaceLastTick = false
```
beanbag44: small thing but i'd call this placedLastTick

src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
```diff
@@ -17,28 +17,79 @@
...
 		listen<MovementEvent.InputUpdate> { event ->
-			event.input.update(forward = 1.0)
-			if (limitSpeed) event.input.movementVector = Vec2f(event.input.strafe, event.input.forward).normalize().multiply(speed.toFloat())
+			val placedLastTick = didPlaceLastTick
+			didPlaceLastTick = false
+
+			if (pauseIfPlacedLastTick && placedLastTick ||
+				pauseWhileMining && breakQueued ||
+				pauseWhilePlacing && placeQueued
+			) return@listen
```
beanbag44, line 85: im always a bit conflicted on what to do with formatting like this. The auto formatting gives this another tab and i never know if its better to put this on the line above at the end. Thats typically what i do tho, but it doesnt really matter

```kotlin
pauseWhilePlacing && placeQueued) return@listen
```

src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
```diff
@@ -17,28 +17,79 @@
...
+
+			val input = event.input
+			if (walkForward || walkBackward) input.forward = (walkForward.toDouble() - walkBackward.toDouble()).toFloat()
```
beanbag44: you can replace the toDoubles here with toFloats which would remove the need to toFloats at the end

src/main/kotlin/com/lambda/module/modules/movement/AutoWalk.kt
```diff
@@ -17,28 +17,79 @@
...
+			if (strafeLeft || strafeRight) input.strafe = (strafeRight.toDouble() - strafeLeft.toDouble()).toFloat()
+			if (sneak || sprint) input.update(sneak = sneak || input.sneaking, sprint = sprint || input.sprinting)
+
+			if (limitSpeed) input.movementVector = Vec2f(input.strafe, input.forward).normalize().multiply(speed.toFloat())
```
beanbag44: as speed it only used once here and it is cast to a float, its best just to make the speed setting a float setting

beanbag44 (changes requested):
The module seems to flag a lot for me on 2b if i have sprint disabled and movement is directed in any direction but forward. The left and right are reversed and i believe the didPlaceLastTick only catches every other tick. If you hold down right click while looking down, it will only slow you down rather than making you stop. The module as a whole also seems to flag when colliding with walls. Not sure if the wall flagging was there before or after the pr changes though.

PoTheMagicDragon:
> The module seems to flag a lot for me on 2b if i have sprint disabled and movement is directed in any direction but forward. The left and right are reversed and i believe the didPlaceLastTick only catches every other tick. If you hold down right click while looking down, it will only slow you down rather than making you stop. The module as a whole also seems to flag when colliding with walls. Not sure if the wall flagging was there before or after the pr changes though.

I'll look into the flagging soon, but I think the `placedLastTick` is working correctly. The default place speed in minecraft when holding down right click is to place a block every 4 ticks. I suppose I could add a `ticksToWaitAfterPlacing` setting. Thoughts on that?

beanbag44:
> I'll look into the flagging soon, but I think the `placedLastTick` is working correctly. The default place speed in minecraft when holding down right click is to place a block every 4 ticks. I suppose I could add a `ticksToWaitAfterPlacing` setting. Thoughts on that?

The ticksToWaitAfterPlacing setting sounds good either way. But yeah i just tested and it is every 4 ticks or whatever the delay is. Strange, i remember packet logging it at one point to see what was going on with something i wrote and it looked much faster. I suppose something could have changed between versions or its just always been that way

PoTheMagicDragon:
Ready for review again. Note, I removed limitSpeed/ Speed because any sideways movement whatsoever (bumped by a mob, hit a wall, tapped left/right) instantly flags you. It can still technically work, but I don't think it's worth including

PR #323, opened 2026-08-19
Target strafe rewrite, the old one was not working

KMatias123:
target strafe working :>

the code is [this](https://github.com/KMatias123/targetstrafe-lambda/) ported to the modern lambda base

This commit also modified the ClientPlayerEntityMixin as this needed to be able to set the x and z values

Tested on test.ccbluex.net and it bypasses their updated-ncp, (3.17.1-SNAPSHOT-Updated-b198) (latest build would be 279 so ig that is quite outdated)

### DIFF
--- a/src/main/kotlin/com/lambda/event/events/MovementEvent.kt
+++ b/src/main/kotlin/com/lambda/event/events/MovementEvent.kt
@@ -25,7 +25,7 @@ import net.minecraft.entity.LivingEntity
 import net.minecraft.entity.MovementType
 import net.minecraft.util.math.Vec3d
 
-sealed class MovementEvent {
+abstract class MovementEvent {
     /**
      * Represents player movement update events.
      * This event will even be triggered if the player is not moving.
@@ -37,18 +37,18 @@ sealed class MovementEvent {
         /**
          * Event triggered before player movement.
          *
-         * @property movementType The type of movement.
+         * @property movement Type The type of movement.
          * @property movement The movement vector.
          */
         data class Pre(
             override val movementType: MovementType,
-            override val movement: Vec3d,
+            override var movement: Vec3d,
         ) : Player(), ICancellable by Cancellable()
 
         /**
          * Event triggered after player movement.
          *
-         * @property movementType The type of movement.
+         * @property movement Type The type of movement.
          * @property movement The movement vector.
          */
         data class Post(
--- a/src/main/kotlin/com/lambda/event/events/MovementEvent.kt
+++ b/src/main/kotlin/com/lambda/event/events/MovementEvent.kt
@@ -25,7 +25,7 @@ import net.minecraft.entity.LivingEntity
 import net.minecraft.entity.MovementType
 import net.minecraft.util.math.Vec3d
 
-abstract class MovementEvent {
+sealed class MovementEvent {
     /**
      * Represents player movement update events.
      * This event will even be triggered if the player is not moving.
--- a/src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
+++ b/src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
@@ -17,18 +17,22 @@
 
 package com.lambda.module.modules.movement
 
+import com.lambda.config.ConfigEditor.forEachSetting
+import com.lambda.config.ConfigEditor.hideAllExcept
+import com.lambda.config.blocks.WorldLineSettings
+import com.lambda.config.withEdits
 import com.lambda.context.SafeContext
 import com.lambda.event.events.MovementEvent
-import com.lambda.event.events.RenderEvent
 import com.lambda.event.events.TickEvent
 import com.lambda.event.listener.SafeListener.Companion.listen
-import com.lambda.interaction.managers.rotating.Rotation
+import com.lambda.graphics.mc.renderer.ImmediateRenderer.Companion.immediateRenderer
 import com.lambda.interaction.managers.rotating.Rotation.Companion.rotationTo
 import com.lambda.module.Module
 import com.lambda.module.modules.combat.KillAura
 import com.lambda.module.tag.ModuleTag
 import net.minecraft.entity.effect.StatusEffects
 import net.minecraft.util.math.Vec3d
+import java.awt.Color
 import kotlin.math.cos
 import kotlin.math.sin
 import kotlin.math.sqrt
@@ -47,8 +51,19 @@ object TargetStrafe : Module(
     private val needsAura by setting("NeedsAura", true)
     private val antiStuck by setting("AntiStuck", true)
 
-    // private val renderCircle by setting("RenderCircle", true)
-    // private val renderThickness by setting("RenderThickness", 2f, 0.5f..8f, 0.5f, visibility = { renderCircle })
+    private val renderCircle by setting("RenderCircle", true)
+    private val renderCircleColor by setting("RenderCircleColor", Color(255, 255, 255, 100), visibility = { renderCircle })
+    private val renderThickness by configBlock(WorldLineSettings(this))
+         .withEdits {
+             hideAllExcept(
+                 ::distanceScaling,
+                 ::worldWidthSetting,
+                 ::screenWidthSetting
+             )
+             forEachSetting {
+                 visibility { old -> { old() && renderCircle } }
+             }
+         }
 
     private var direction = 1
 
@@ -59,22 +74,19 @@ object TargetStrafe : Module(
 
     init {
         listen<TickEvent.Post> {
-
             if (strafing && autoJump && player.isOnGround) {
                 player.jump()
             }
         }
 
         listen<MovementEvent.Player.Pre> { event ->
-            if (mc.player!!.horizontalCollision && antiStuck) {
+            if (player.horizontalCollision && antiStuck) {
                 switchDirection()
             }
-            val strafe = canStrafe()
-
-            if (strafe) {
-                val rotations = KillAura.target?.let { it1 -> player.eyePos!!.rotationTo(it1.pos) } ?: return@listen
+            if (canStrafe()) {
+                val rotations = KillAura.target?.let { it1 -> player.eyePos?.rotationTo(it1.pos) } ?: return@listen
                 KillAura.target?.let { it1 -> doStrafeAtSpeed(event, rotations.yawF, it1.pos) }
-                val r: Rotation
+                currentTargetVec = KillAura.target?.pos
 
                 strafing = true
             } else {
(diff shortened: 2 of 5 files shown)


src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,192 @@
...
+    // private val renderThickness by setting("RenderThickness", 2f, 0.5f..8f, 0.5f, visibility = { renderCircle })
 
-    private var forwardDirection = 1.0
-    private var strafeDirection = 1.0
+    private var direction = 1
 
-    @JvmStatic
-    val isActive get() = isEnabled && KillAura.isEnabled && KillAura.target != null
+    private var currentDistance = 0.toDouble()
+    private var currentTargetVec: Vec3d? = null
```
beanbag44, line 71: this doesnt seem to be used anywhere

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,192 @@
...
-            if (KillAura.target == null) {
-                forwardDirection = 1.0
-                strafeDirection = 1.0
+            if (strafing && autoJump && player.isOnGround) {
+                player.jump()
+            }
+        }
+
+        listen<MovementEvent.Player.Pre> { event ->
+            if (mc.player!!.horizontalCollision && antiStuck) {
```
beanbag44: this listener is a safe listener, so you should be able to reference player directly to avoid the !! and mc receiver

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,192 @@
...
+                switchDirection()
+            }
+            val strafe = canStrafe()
+
+            if (strafe) {
+                val rotations = KillAura.target?.let { it1 -> player.eyePos!!.rotationTo(it1.pos) } ?: return@listen
+                KillAura.target?.let { it1 -> doStrafeAtSpeed(event, rotations.yawF, it1.pos) }
+                val r: Rotation
```
beanbag44: r doesnt seem to be used either

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,192 @@
...
+        }
+
+        listen<MovementEvent.Player.Pre> { event ->
+            if (mc.player!!.horizontalCollision && antiStuck) {
+                switchDirection()
+            }
+            val strafe = canStrafe()
+
+            if (strafe) {
+                val rotations = KillAura.target?.let { it1 -> player.eyePos!!.rotationTo(it1.pos) } ?: return@listen
```
beanbag44: shouldnt need the !! here after eyePos

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,192 @@
...
+            if (strafing && autoJump && player.isOnGround) {
+                player.jump()
+            }
+        }
+
+        listen<MovementEvent.Player.Pre> { event ->
+            if (mc.player!!.horizontalCollision && antiStuck) {
+                switchDirection()
+            }
+            val strafe = canStrafe()
```
beanbag44: could be worth setting strafing here instead of making a local strafe value and copying it to strafing later on

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,192 @@
...
-            forwardDirection = 1.0
-            strafeDirection = 1.0
+        currentDistance = distance
+
+
+        // jump boost
+
+        val jumpboost = player.getStatusEffect(StatusEffects.JUMP_BOOST)
+        if (jumpboost != null) {
+            jumpVelocity *= jumpboost.amplifier
```
beanbag44: unused

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,192 @@
...
+        }
+
+        event.movement = Vec3d(playerSpeed * cos(Math.toRadians((rotationYaw + 90.0f).toDouble())), event.movement.y, playerSpeed * sin(Math.toRadians((rotationYaw + 90.0f).toDouble())))
+
+
+        return false
+    }
+
+    private fun canStrafe(): Boolean {
+        if ((KillAura.isEnabled || !needsAura) && KillAura.target != null) {
```
beanbag44: could simplify to remove the if statement as a whole

beanbag44 (changes requested):
Havent tested yet but will in a bit. The defaults for lambda at the moment are usually for grim/2b2t's anticheat but its ok for now as i assume this is an improvement either way. Theres a lotta blank spaces throughout the module which could be cleaned up and if you want to, we have the rendering api in place which can draw circles too! Just for future reference, i almost never use !! as its an unsafe operator which could cause null pointer exceptions. Always best to use ?. checks and return if null or something like that

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: removed if

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: removed

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: inlined

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: added ?

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: i was editing in emacs and wanted to import Rotation so it was easy to write that and then make the language server import Rotation, now it is removed

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: done

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: was used in rendering but i forgot to comment it out, now it is used again in rendering

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
beanbag44: i think eyePos is always non null so the ? isnt needed either afaik but theres no issue with this as the null is checked anyway

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,134 @@
...
+    private val distanceSetting by setting("PreferredDistance", 1f, 0f..6f, 0.1f)
+    private val maxDistance by setting("MaxDistance", 10f, 1f..32f, 0.5f)
+    private val turnAmount by setting("TurnAmount", 5f, 1f..90f, 0.5f)
+    private val hSpeed by setting("HSpeed", 0.2873f, 0.001f..10.0f, 0.0001f)
+
+    private val needsAura by setting("NeedsAura", true)
+    private val antiStuck by setting("AntiStuck", true)
+
+    private val renderCircle by setting("RenderCircle", true)
+    private val renderCircleColor by setting("RenderCircleColor", Color(255, 255, 255, 100), visibility = { renderCircle })
```
beanbag44: the visibility can be moved out of the parenthesis here

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
```diff
@@ -17,80 +17,134 @@
...
+        if (speed != null) {
+
+            playerSpeed *= 1.0f + 0.2f * (speed.amplifier + 1)
         }
+
+        event.movement = Vec3d(playerSpeed * cos(Math.toRadians((rotationYaw + 90.0f).toDouble())), event.movement.y, playerSpeed * sin(Math.toRadians((rotationYaw + 90.0f).toDouble())))
+        return false
+    }
+
+    private fun canStrafe(): Boolean {
```
beanbag44, line 143: can remove the return type, brackets, and return keyword to replace with a simple = after the (), with just the return after on the same line. A small thing but thought i'd mention anyway

src/main/kotlin/com/lambda/event/events/MovementEvent.kt
```diff
@@ -25,7 +25,7 @@ import net.minecraft.entity.LivingEntity
 import net.minecraft.entity.MovementType
 import net.minecraft.util.math.Vec3d
 
-sealed class MovementEvent {
+abstract class MovementEvent {
```
beanbag44: why change this to abstract?

beanbag44 (changes requested):
pretty much good to go, just a couple small things

beanbag44:
oh also private var currentDistance = 0.toDouble() can just use 0.0 instead of 0.toDouble()

src/main/kotlin/com/lambda/event/events/MovementEvent.kt
KMatias123: i first played with the class types when planning on what to change but ended up just making those 2 vars instead of vals, but then forgot about that single keyword when i reverted the stuff i changed, could have used git revert

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: i can change it if you want but imo using readable syntax is important for an open source project

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: how? it needs a lambda expression?

beanbag44: in kotlin you can do ...255, 255, 100)) { renderCircle }
because its at the end of the parameters

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
beanbag44: i consider it more readable but its fine not changing it. Just less verbose

KMatias123: didn't change as this is quite unique syntax if a user wants to read the code

src/main/kotlin/com/lambda/module/modules/movement/TargetStrafe.kt
KMatias123: changed

PR #326, opened 2026-08-20
Improve crystal aura (second try at git)

KMatias123:
note: this has **not** been tested on 2b2t as i do not have priority queue, i have tested it on ccbluex's server

additions:
* rendering and settings
* the hand selected in the settings is now what hand the module prefers to use, so if you have crystals in your offhand while having the hand setting set to main and need to place some it will not start to move crystals from the inventory and instead places the ones in the inventory, after they are used it will then start moving crystals to the hotbar

fixes:
* trying to place at opportunities that are blocked by other crystals
* main hand swapping when main hand is selected as preference
* hotbar logic

improvements:
* made it bypass grim, also with the prediction exploit

### DIFF
--- a/src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
+++ b/src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
@@ -45,18 +45,15 @@ import com.lambda.threading.runSafeAutomated
 import com.lambda.threading.runSafeGameScheduled
 import com.lambda.util.BlockUtils.blockState
 import com.lambda.util.CommunicationUtils.info
-import com.lambda.util.CommunicationUtils.log
 import com.lambda.util.PacketUtils.sendPacket
 import com.lambda.util.Timer
 import com.lambda.util.collections.LimitedDecayQueue
 import com.lambda.util.combat.CombatUtils.crystalDamage
 import com.lambda.util.extension.fullHealth
-import com.lambda.util.item.ItemStackUtils.slotId
 import com.lambda.util.math.*
 import com.lambda.util.math.MathUtils.ceilToInt
 import com.lambda.util.math.MathUtils.roundToStep
 import com.lambda.util.player.RotationUtils.getVisibleSurfaces
-import com.lambda.util.player.SlotUtils.hotbarSlots
 import com.lambda.util.player.SlotUtils.hotbarStacks
 import com.lambda.util.world.fastEntitySearch
 import net.minecraft.block.Blocks
@@ -92,7 +89,7 @@ object CrystalAura : Module(
     private const val RENDERING_TAB = "Rendering"
 
     @Tab(GENERAL_TAB) private val rotate by setting("Rotate", true)
-    @Tab(GENERAL_TAB) private val updateMode by setting("Update Mode", UpdateMode.Async)
+    @Tab(GENERAL_TAB) private val updateMode by setting("Update Mode", UpdateMode.Ticked)
     @Tab(GENERAL_TAB) private val updateDelaySetting by setting("Update Delay", 25L, 5L..200L, 5L, unit = " ms") { updateMode == UpdateMode.Async }
     @Tab(GENERAL_TAB) private val maxUpdatesPerFrame by setting("Max Updates Per Frame", 5, 1..20, 1) { updateMode == UpdateMode.Async }
     @Tab(GENERAL_TAB) private val updateDelay get() = if (updateMode == UpdateMode.Async) updateDelaySetting else 0L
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -109,13 +111,32 @@ object CrystalAura : Module(
...
+    @Tab(RENDERING_TAB) private val render by setting("Rendering", true)
+    @Tab(RENDERING_TAB) private val renderPlacements by setting("Placement Rendering", true)
+
+    @Tab(RENDERING_TAB) private val renderLineSettings by configBlock(WorldLineSettings(this))
+        .withEdits {
+            hideAllExcept (
+                ::worldWidthSetting
+            )
+            forEachSetting {
+                visibility { old -> { old() && render }}
```
beanbag44, line 132: we typically keep spaces between }} in scenarios like this. No biggie, just thought i'd say anyway

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -109,13 +111,32 @@ object CrystalAura : Module(
...
 
     @Tab(PREDICTION_TAB) private val targetingSettings by configBlock(TargetingSettings.CombatSettings(this, 10.0))
 
+    @Tab(RENDERING_TAB) private val render by setting("Rendering", true)
+    @Tab(RENDERING_TAB) private val renderPlacements by setting("Placement Rendering", true)
+
+    @Tab(RENDERING_TAB) private val renderLineSettings by configBlock(WorldLineSettings(this))
+        .withEdits {
+            hideAllExcept (
+                ::worldWidthSetting
```
beanbag44, line 129: can inline this so its (::worldWidthSetting)

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -240,6 +268,24 @@ object CrystalAura : Module(
             decay += crystal.id
         }
 
+        immediateRenderer("CrystalAura Immediate Renderer") {
+            runSafe {
+                if (lastPlace != null) {
+                    if (lastPlace!!.second + 100 < System.currentTimeMillis()) {
```
beanbag44, line 274: we never use !! as its an unsafe operator. Could be changed to use

val place = lastPlace ?: return@listen

then using the place value as we know its not null

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -285,14 +330,19 @@ object CrystalAura : Module(
 	}
 
 	private fun SafeContext.placeInternal(opportunity: Opportunity, hand: Hand) {
-		interaction.syncSelectedSlot()
-		connection.sendPacket {
-			PlayerInteractBlockC2SPacket(
-				hand, BlockHitResult(opportunity.crystalPosition, opportunity.side, opportunity.blockPos, false), 0
-			)
-		}
+        runSafe {
```
beanbag44, line 333: we're already in a SafeContext so we shouldnt need runSafe here

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -285,14 +330,19 @@ object CrystalAura : Module(
...
+            lastPlace = Pair(opportunity.blockPos, System.currentTimeMillis())
+        }
+		interaction.syncSelectedSlot() // TODO: when server only hotbar swap gets implemented, this will be removed
+        mc.interactionManager!!.sendSequencedPacket(world, { sequence ->
```
beanbag44, line 337: can just use interaction.sendSequencedPacket due to the SafeContext

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -285,14 +330,19 @@ object CrystalAura : Module(
...
+
+            PlayerInteractBlockC2SPacket(
+                hand, BlockHitResult(opportunity.crystalPosition, opportunity.side, opportunity.blockPos, false), sequence
+            )
+        });
```
beanbag44, line 342: no need for ; here and the lambda can be moved out of the parenthesis like how the settings do visibility

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -361,7 +411,7 @@ object CrystalAura : Module(
 
                 val entitiesNearby = fastEntitySearch<Entity>(3.5, pos)
                 val crystals = entitiesNearby.filterIsInstance<EndCrystalEntity>()
-                val otherEntities = entitiesNearby - crystals + player
+                val otherEntities = entitiesNearby - crystals.toSet() + player
```
beanbag44, line 414: is this to make duplicate lookups faster?

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -481,38 +530,56 @@ object CrystalAura : Module(
...
+                } else null
+
+                val swapTo = when (swapHand) {
+                    Hand.MAIN_HAND -> HotbarContainer
+                    Hand.OFF_HAND -> OffHandContainer
+                }
+
+                if (itemToUse == null || swapHand == Hand.MAIN_HAND) {
+                    if (swapHand == Hand.MAIN_HAND) {
+                        val itemStack = selection.bestItemMatch(player.hotbarStacks)
```
beanbag44, line 554: this works, but ive been migrating things to use the containers. This would instead use HotbarContainer to access the slots

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -481,38 +530,56 @@ object CrystalAura : Module(
...
+                        if (itemStack == null) { // retrieve to hotbar
+                            if (!selection.transfer(swapTo)) return@runSafe
+                        }
+                        val s = player.hotbarStacks.indexOf(itemStack)
```
beanbag44, line 558: instead of indexOf here, you could use the container as said above, and then access the index through Slot::index

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
```diff
@@ -481,38 +530,56 @@ object CrystalAura : Module(
...
+                    } else { // retrieve to offhand
+                        if (!selection.transfer(swapTo)) return@runSafe
+                    }
+                }
 			}
 
-            placeTimer.runSafeIfPassed(placeDelay.milliseconds) {
-                placeInternal(this@Opportunity, swapHand)
+            if (placeTimer.timePassed(placeDelay.milliseconds) || (postPacketPlace && safeToPlaceInstantly && !placePostPause)) {
+                runSafe {
```
beanbag44, line 567: shouldnt be a need here for runSafe as we're already in a runSafe context from the start of the function

beanbag44 (changes requested):

src/main/kotlin/com/lambda/module/modules/combat/CrystalAura.kt
KMatias123: yeah

PR #330, opened 2026-09-15
adds stripped state pre processor

PoTheMagicDragon:
### Description
Adds a pre processor for handling stripped logs


src/main/kotlin/com/lambda/interaction/construction/simulation/processing/PlacementProcessor.kt
```diff
@@ -27,6 +28,10 @@ import net.minecraft.util.math.BlockPos
  * unnecessary to scan all of them, for example.
  */
 interface StateProcessor {
+	/** Checked before [acceptsState], so a disabled processor still falls back to breaking. */
+	context(_: AutomatedSafeContext)
+	fun isEnabled(): Boolean = true
```
beanbag44, line 33: the isEnabled check is only overridden by StrippedStateProcessor so the check should probably be moved into the acceptsState for StrippedStateProcessor and the function removed.

src/main/kotlin/com/lambda/interaction/construction/simulation/processing/preprocessors/state/StrippedStateProcessor.kt
```diff
@@ -0,0 +1,88 @@
...
+	// [AxeItem.STRIPPED_BLOCKS] maps unstripped to stripped, we need the other direction
+	private val unstrippedToStripped: Map<Block, Block> by lazy {
+		AxeItem.STRIPPED_BLOCKS.entries.associate { (from, to) -> to to from }
+	}
+
+	context(automatedSafeContext: AutomatedSafeContext)
+	override fun isEnabled() = automatedSafeContext.buildConfig.stripLogs
+
+	override fun acceptsState(state: BlockState, targetState: BlockState): Boolean {
+		val unstrippedVariant = unstrippedToStripped[targetState.block] ?: return false
```
beanbag44, line 44: this function as a whole can probably be simplified to an = function without a full block

src/main/kotlin/com/lambda/interaction/construction/simulation/processing/preprocessors/state/StrippedStateProcessor.kt
```diff
@@ -0,0 +1,88 @@
...
+		return state.isReplaceable || state.block == unstrippedVariant
```
beanbag44, line 45: state.isReplaceable shouldnt be necessary here. Its best to have strict requirements for a preprocessor to accept. If the state is replaceable, the preprocessor doesnt need to pre process as its not a log

src/main/kotlin/com/lambda/interaction/construction/simulation/processing/ProcessorRegistry.kt
```diff
@@ -144,15 +143,19 @@ object ProcessorRegistry : Loadable {
...
-	context(safeContext: SafeContext)
+	context(_: AutomatedSafeContext)
 	private fun preProcess(pos: BlockPos, state: BlockState, targetState: BlockState, itemStack: ItemStack) =
 		PreProcessingInfoAccumulator(targetState, itemStack.item).run {
 			var stateProcessing = false
 			stateProcessors.forEach { processor ->
-				if (processor.acceptsState(state, targetState)) {
-					with(processor) { preProcess(state, targetState, pos) }
-					stateProcessing = true
+				if (!processor.acceptsState(state, targetState)) return@forEach
```
beanbag44, line 151: this shouldnt be altered

beanbag44 (changes requested):
small stuff

src/main/kotlin/com/lambda/interaction/construction/simulation/processing/preprocessors/state/StrippedStateProcessor.kt
PoTheMagicDragon: How do we make it place the unstripped log in the correct spot then?

src/main/kotlin/com/lambda/interaction/construction/simulation/processing/preprocessors/state/StrippedStateProcessor.kt
```diff
@@ -0,0 +1,88 @@
...
+
+	context(safeContext: SafeContext)
+	override fun PreProcessingInfoAccumulator.preProcess(state: BlockState, targetState: BlockState, pos: BlockPos) {
+		val unstrippedVariant = unstrippedToStripped[targetState.block] ?: return
+		val player = safeContext.player
+
+		// Every branch below depends on the current inventory, so this must never be cached.
+		noCaching()
+
+		if (state.isReplaceable){
```
beanbag44, line 56: missing space before the {

Also, if you want the logic to mirror FlowerPotStateProcessor for parity, you could change this to state.block != unstrippedVariant and add a return in the block so you can handle the stripping logic after without indentation. The isReplaceable check is already handled in the acceptsState function

src/main/kotlin/com/lambda/interaction/construction/simulation/processing/preprocessors/state/StrippedStateProcessor.kt
beanbag44: hm actually yeah, sorry i was mistaken. I should have checked the FlowerPotStateProcessor, the transformation mirrors this

src/main/kotlin/com/lambda/interaction/construction/simulation/processing/preprocessors/state/StrippedStateProcessor.kt
```diff
@@ -0,0 +1,88 @@
...
+
+	override fun acceptsState(state: BlockState, targetState: BlockState): Boolean {
+		val unstrippedVariant = unstrippedToStripped[targetState.block] ?: return false
+		return state.isReplaceable || state.block == unstrippedVariant
+	}
+
+	context(safeContext: SafeContext)
+	override fun PreProcessingInfoAccumulator.preProcess(state: BlockState, targetState: BlockState, pos: BlockPos) {
+		val unstrippedVariant = unstrippedToStripped[targetState.block] ?: return
+		val player = safeContext.player
```
beanbag44, line 51: for something like this, i try to avoid value sets and call with(safeContext) {  }, wrapping the function in the safe context. If you change it to use that, i would change the function to = without the {}  and then put the with on the next line

beanbag44 (changes requested):
yeah, i was mistaken, sorry. Added a few other small things but theyre not important (aside from the missing space before the opening bracket)

PR #331, opened 2026-09-17
smart pitch for bounce for 48bps

sophimoo:
When enabled you can get up to 48bps (sometimes 50) when bouncing compared to the normal 40 bps at 72 degrees


sophimoo:
okay idk which version i like better
```kt
	val targetPitch = when {
		!smartPitch -> pitch
		player.velocity.y > -0.2 -> 90.0
		else -> 4.0
	}
	rotationRequest { pitch(targetPitch) }.submit()
```

or 

```kt
	if (smartPitch) rotationRequest { pitch(if (player.velocity.y > -0.2) 90.0 else 4.0) }.submit()
	else rotationRequest { pitch(pitch) }.submit()

```

The second one reads more readable to me but the first avoids doing submit twice

beanbag44:
I prefer the first one, but you could also do:
```kotlin
rotationRequest {
    when {
        !smartPitch -> pitch(pitch)
        player.velocity.y > -0.2 -> pitch(90.0)
        else -> pitch(4.0)
    }
}.submit()
```
which would be my pick.

Also, could the pitch setting be hidden if Smart is true? And if so, the Smart setting should probably be above the Pitch setting

